PR #14, opened 2020-07-03
fix math transforms involving an undefined function

lkeegan:
## Description
- add check that function exists before calling replaceFD()
- if function doesn't exist, return NaN
- add tests

## Motivation and Context
- calling `evaluateASTNode` with an AST involving an undefined function would crash
- this was caused by an infinite `replaceFD()` recursion ending in a stack-overflow error

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/SBMLTransforms.cpp
+++ b/src/sbml/SBMLTransforms.cpp
@@ -652,12 +652,24 @@ SBMLTransforms::evaluateASTNode(const ASTNode * node, const IdValueMap& values, 
   case AST_FUNCTION:
     /* shouldnt get here */
     // but we do if math we are expanding uses a functionDefinition
-    lfd = m->getListOfFunctionDefinitions();
-    tempNode = node->deepCopy();
-    replaceFD(tempNode, lfd);
-    result = evaluateASTNode(tempNode, values, m);
-    delete tempNode;
-    break;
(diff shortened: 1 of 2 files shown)


src/sbml/SBMLTransforms.cpp
```diff
@@ -652,12 +652,35 @@ SBMLTransforms::evaluateASTNode(const ASTNode * node, const IdValueMap& values,
...
-    break;
+    {
+        if(m != NULL){
+            lfd = m->getListOfFunctionDefinitions();
+        }
+        unsigned nFunctionDefinitions = 0;
+        if (lfd != NULL){
+            nFunctionDefinitions = lfd->size();
+        }
+        bool functionDefinitionExists{false};
```
fbergmann: here i think it would be cleaner to just use the `get(sid)` overload of the list of functiondefinitions, something like 


     bool functionDefinitionExists = lfd->get(node->getName()) != NULL;


would that work for you? (in general usually all libsbml getters exist once for index, and once for id).

lkeegan: yes that's much cleaner - thanks, I'll update the PR

fbergmann (approved):
this looks perfect, thanks again

PR #17, opened 2020-07-03
 - #11 : behavior of reading out of the allocated range is undefined,…

### LINKED ISSUE #11: spatial tests: test_SampledField_samples_mismatchLength_1
In `test_SampledField_samples_mismatchLength_1`, this test condition fails for me:

https://github.com/sbmlteam/libsbml/blob/5744b9c3788efb06584cf7fb0a7dc6109d1c12ac/src/sbml/packages/spatial/sbml/test/TestImageData.cpp#L177

This element was initialized from the third element of an array of size 2, so its value is whatever was at that address, which could be anything..

including zero, in which case the test fails.

fbergmann:
… so removing these tests

## Description
as discussed in issue #11 , the behavior of accessing elements past the declared range is undefined, so these tests may or may not fail. 

## Motivation and Context
 - the test was not actually testing what it wanted to test. 

fixes #11 

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/packages/spatial/sbml/test/TestImageData.cpp
+++ b/src/sbml/packages/spatial/sbml/test/TestImageData.cpp
@@ -156,63 +156,6 @@ START_TEST (test_SampledField_samples)
 END_TEST
 
 
-START_TEST (test_SampledField_samples_mismatchLength_1)
-{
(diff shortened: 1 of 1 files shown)


skeating (approved):
Okay I'll look at this as code review of spatial

PR #18, opened 2020-07-03
issue #2 : ensure that configure script is correctly updating

### LINKED ISSUE #2: R configuration and build
These need to be fixed for the old stable vs experimental issues.

fbergmann:
- the configure script will now automatically update based on what packages are enabled
- it is now possible to just create an R source package by defining WITH_CREATE_R_SOURCE and WITH_SKIP_R_BINARY

## Description
- modified cmakelists.txt to add the flag `WITH_SKIP_R_BINARY`, if that is set, the R binary bindings are not created, that way it is now possible to just create the R source package on windows
- the cmake configuration now correctly updates the list of packages needed in the configure scripts of the source archives. 

## Motivation and Context
 - when a source archive is created with current branch, it fails as it is missing the math package, by once and for all generating the right flags, this will sort things out
 - fixes #2 

## Types of changes

## Checklist:

## Testing
- [] Testing is done automatically and codecov shows test coverage
- [X] This cannot be tested automatically, we ought to add a step to travis later, for now, configure with_skip_r_binary and create source r package, create the source archive and run R CMD check on the source archive (just did and it worked for me)


fbergmann:
To make this easier to check, here is how to create the source archive

```
  git clone https://github.com/sbmlteam/libsbml
  git checkout fix-r-source-archive
  mkdir build_sbml 
  cd build_sbml
  cmake -DWITH_R=ON -DWITH_CREATE_R_SOURCE=ON -DWITH_SKIP_R_BINARY=ON ../libsbml
  make binding_r_source_package
  R CMD check src/bindings/r/out/libSBML_5.18.3.tar.gz

```

the last step takes forever, as compiling everything is a pain, but in the end you get the output: 

* checking whether package ‘libSBML’ can be installed ... OK

(later on it continues to fail because of the PDF thing, but that is a known issue that was always there)

skeating (approved):

PR #20, opened 2020-07-03
 issue #19: prevent delete of released memory

### LINKED ISSUE #19: sort potential memory leaks in unitformatter
Sonarcloud lists quite the number of blocking issues with the unit formatter, looking into those: 

<https://sonarcloud.io/project/issues?id=sbmlteam_libsbml>

fbergmann:
## Description
 - sonar can't detect that the loop is only entered once, thus added break statement
 - ensure that child1 and child2 are set to NULL, as static analysis found a path, in which memory could be deleted twice

## Motivation and Context
 - freeeing released memory results in undefined behavior whereas deleting NULL is safe.
fixes #19 

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/units/UnitFormulaFormatter.cpp
+++ b/src/sbml/units/UnitFormulaFormatter.cpp
@@ -1663,7 +1663,7 @@ UnitFormulaFormatter::getUnitDefinitionFromSpecies(const Species * species)
     {
       for (n = 0; n < model->getNumUnitDefinitions(); n++)
       {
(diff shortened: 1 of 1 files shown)


skeating (approved):
So we need to decide if when we do any refactoring we want sonar to pass all checks.

fbergmann:
i think we merge for now, since it does remove the blockers, and once you have time it is a good one for refactoring

PR #24, opened 2020-07-08
fix memory leak and add test

lkeegan:
## Description
- TestSBMLConvert.c: check that Exponent, Scale and Multiplier are set to default values on upgrade to level 3 if not explicitly set
- Model.cpp: clear Events list in destructor
- TestAttributeFunctions.cpp: delete allocated memory

## Motivation and Context
Fix memory leak & add test to ensure that level 2 units with optional values that and are not explicitly set become explicitly set to their default level 2 values on upgrade to level 3

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/Model.cpp
+++ b/src/sbml/Model.cpp
@@ -159,7 +159,7 @@ Model::~Model ()
     }
     delete mFormulaUnitsData;
   }
-
+  mEvents.clear();
   mUnitsDataMap.clear();
 }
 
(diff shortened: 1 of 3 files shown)


lkeegan:
Sorry, the change to SBMLConvert.cpp is not required - the nan I was getting was due to a locale issue with `strtod`.

I'll remove it, then the PR will just be a couple of memory leak fixes & tests

fbergmann (approved):
thank you again, great catches

PR #25, opened 2020-07-09
use "C" locale version of strtod

lkeegan:
## Description
- Unit.cpp:
  - use `c_locale_strtod()` instead of `strtod()`
  - refactor code fragment used in many places to truncate double precision into `truncateDoublePrecision()`
- TestUnit.c: add test of `Unit::merge()` using "de_DE.UTF-8" locale

## Motivation and Context
If the locale is e.g. "de_DE.UTF-8", then upgrade of a level 2 model to level 3 can produce `nan` in the `Multiplier` of a `Unit` (and hence the upgrade fails due to inconsistent units). This is due to `strtod` being locale dependent - for example in this locale the decimal separator is "," and the string "0.1" is parsed as the number 0.

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/Unit.cpp
+++ b/src/sbml/Unit.cpp
@@ -1526,6 +1526,15 @@ Unit::areEquivalent(Unit * unit1, Unit * unit2)
   return equivalent;
 }
 
+static double truncateDoublePrecision(double x, int num_digits=15){
+    /* hack to force x to be a double with at most
+     * num_digits of precision when written in decimal */
+    std::ostringstream oss;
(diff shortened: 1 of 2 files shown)


fbergmann (commented):
I am wondering whether 'c_locale_strtod' will be available everywhere. Since we are using stringstrems already, could we not read it out of the stream again, after going to the beginning?

<edit> nevermind, just saw that it is one of our methods in util.

fbergmann (approved):
thanks that works for me

PR #46, opened 2020-09-18
Set other 'remove' functions to %newobject in Swig.

luciansmith:
Went through and searched for all non-package '::remove' functions that returned non-owning pointers; added them all to libsbml.i

Probably still need to do this for the packages, though hopefully not for the autogenerated ones.

Did not add tests for this; they'd necessarily be valgrind-type tests.

### DIFF
--- a/src/bindings/swig/comp-package.i
+++ b/src/bindings/swig/comp-package.i
@@ -36,6 +36,15 @@
 
 #ifdef USE_COMP
 
+%newobject removeSubmodel;
+%newobject removePort;
+%newobject removeReplacedElement;
+%newobject removeModelDefinition;
+%newobject removeGeneProduct;
+%newobject removeExternalModelDefinition;
+%newobject removeDeletion;
+
+
 %feature("director") SBMLResolver;  
 
 %ignore Submodel::getAllInstantiatedElements;
--- a/src/bindings/swig/comp-package.i
+++ b/src/bindings/swig/comp-package.i
@@ -40,7 +40,6 @@
 %newobject removePort;
 %newobject removeReplacedElement;
 %newobject removeModelDefinition;
-%newobject removeGeneProduct;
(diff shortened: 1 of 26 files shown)


luciansmith:
This should be ready to merge now, unless someone can think of another class of libsbml functions that return pointers to objects that the caller must delete manually.

src/bindings/swig/comp-package.i
```diff
@@ -36,6 +36,15 @@
 
 #ifdef USE_COMP
 
+%newobject removeSubmodel;
+%newobject removePort;
+%newobject removeReplacedElement;
+%newobject removeModelDefinition;
+%newobject removeGeneProduct;
```
fbergmann: does comp really require a removeGeneProduct?

src/sbml/Constraint.cpp
```diff
@@ -427,7 +427,7 @@ void
 Constraint::replaceSIDWithFunction(const std::string& id, const ASTNode* function)
 {
   if (isSetMath()) {
-    if (mMath->getType() == AST_NAME && mMath->getId() == id) {
+    if (mMath->getType() == AST_NAME && mMath->getName() == id) {
```
fbergmann: it seems like here you have teh same fix included, that are also in the pull request for the comp fix. Are there any differences here? which one should we use?

luciansmith: Bah, sorry--this fix should indeed be part of the other pull request.  Did it get in both?  In the end, this line needs to be 'getName', however that works out.

src/bindings/swig/comp-package.i
luciansmith: Nope!  Good catch.

luciansmith:
OK!  The redundant changes have now been merged away.

fbergmann (approved):
that works for me .. thanks

PR #47, opened 2020-09-29
Fix conversion problem issue 480.

luciansmith:
The situation described in issue 480 (a conversion factor that has to be applied to a top-level ASTNode) was anticipated in the code, but a bug prevented it from triggering:  the ASTNode's 'id' was being checked, instead of the 'name'.

Fixes https://sourceforge.net/p/sbml/libsbml/480/ and adds tests.

### DIFF
--- a/src/sbml/Constraint.cpp
+++ b/src/sbml/Constraint.cpp
@@ -427,7 +427,7 @@ void
 Constraint::replaceSIDWithFunction(const std::string& id, const ASTNode* function)
 {
   if (isSetMath()) {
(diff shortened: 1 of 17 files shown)


fbergmann:
the fix here is also included for the pull request for the swig issue ... so we should only apply one of them.

luciansmith:
Bah, git.

If you approve this one first, I'll merge the changes into the %newobject branch, which should zero out those redundant changes.

fbergmann (approved):
this does look good to me .. thanks!

PR #52, opened 2020-10-12
remove test with undefined behaviour

lkeegan:
## Description

- remove check that value is zero

## Motivation and Context

- the value being tested here was previously (intentionally) copied from beyond the end of an array
- so it could have any value (including zero, in which case this test fails)

## Types of changes

## Checklist:

## Testing


fbergmann (approved):
that works for me .. thanks

PR #53, opened 2020-10-12
add const version of Geometry::getCoordinateComponentByKind

lkeegan:
## Types of changes

## Checklist:

## Testing


fbergmann (commented):
we usually went the other way when adding const variants, that is implementing the non-const version, and then using covariant return types in the overload to get the non-const version converted to const. 

Is there a specific reason to do it the other way around? Just wondering.

lkeegan:
No reason - either way is fine as far as I know. I've swapped them around, does it match what you normally do now?

fbergmann (approved):
yup .. thank you so very much

PR #59, opened 2020-10-21
 - issue #57: return the ListWrapper class, rather than a pointer to …

### LINKED ISSUE #57: Potential Memory Leak in `getListOfAllElements`
https://github.com/AMICI-dev/AMICI/pull/1310/checks?check_run_id=1284234058 suggests that calling `getListOfAllElements` is leaking memory. 

Can be reproduced with the following function: https://github.com/AMICI-dev/AMICI/blob/96df68d11e7fcb6db1bb717a2b5417070182a7d3/python/amici/sbml_import.py#L1830

fbergmann:
…it, which will be freed by the  target language

## Description
In issue #57 , Fabian notices a memory leak that occurs, since we return a pointer to a listwrapper object, rather than the object itself. I've tried to combine the swig %extend syntax with %newobject, to go about without any changes, however, i was not successful in doing so. Comparing the generated code, the result is precisely what we want, we now get statements of form: 

`
  resultobj = SWIG_NewPointerObj((new ListWrapper< SBase >(static_cast< const ListWrapper< SBase >& >(result))), SWIGTYPE_p_ListWrapperT_SBase_t, SWIG_POINTER_OWN |  0 );
`

denoting that the pointer is owned by the target language. 

## Motivation and Context
fixes #57 

## Types of changes

## Checklist:

## Testing
I would feel better if someone else gives this a try, i have only tried python and C# bindings yet.

### DIFF
--- a/src/bindings/swig/comp-package.i
+++ b/src/bindings/swig/comp-package.i
@@ -50,10 +50,10 @@
 
 %extend Submodel
 {
-	ListWrapper<SBase>* getListOfAllInstantiatedElements()
+	ListWrapper<SBase> getListOfAllInstantiatedElements()
 	{
 		List* list = $self->getAllInstantiatedElements();
-		return new ListWrapper<SBase>(list);
+		return ListWrapper<SBase>(list);
 	}
 }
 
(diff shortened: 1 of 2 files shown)


skeating:
@fbergmann should this be a pull request ;-)

fbergmann:
@skeating , indeed it should and it is ... and requires review by someone that is not me 

once approved, i'll release new python bindings with this patch

skeating (approved):

PR #60, opened 2020-10-26
Add the ability to sort ListOf's by SId.

luciansmith:
In my tests of Antimony, I round-trip stuff through SBML, and the order sometimes changes.  I need a function like this to sort the lists so I can do simple before-and-after comparisons.

## Types of changes

- [X ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ X] I have updated all documentation necessary.
- [ X] I have checked spelling in (new) comments.

## Testing

### DIFF
--- a/src/sbml/ListOf.cpp
+++ b/src/sbml/ListOf.cpp
@@ -687,13 +687,30 @@ ListOf::setExplicitlyListed(bool value)
 
 struct ListOfComparator
 {
-    // Compare 2 SBase* objects using name
+    // Compare 2 SBase* objects using the 'id' attribute, 
+    // then the results of 'getId' (which is sometimes different 
+    // from 'getId'), then name, then metaid.
     bool operator ()(const SBase* obj1, const SBase* obj2)
     {
         if (obj1 == NULL || obj2 == NULL) {
             return true;
         }
-        return obj1->getId() < obj2->getId();
+        if (obj1->getIdAttribute() == obj2->getIdAttribute()) {
+            if (obj1->getId() == obj2->getId()) {
+                if (obj1->getName() == obj2->getName()) {
+                    return obj1->getMetaId() < obj2->getMetaId();
+                }
+                else {
+                    return obj1->getName() < obj2->getName();
+                }
+            }
+            else {
+                return obj1->getId() < obj2->getId();
+            }
+        }
+        else {
+            return obj1->getIdAttribute() < obj2->getIdAttribute();
+        }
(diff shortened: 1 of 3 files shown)


src/sbml/ListOf.cpp
```diff
@@ -685,6 +685,23 @@ ListOf::setExplicitlyListed(bool value)
   mExplicitlyListed = value;
 }
 
+struct ListOfComparator
+{
+    // Compare 2 SBase* objects using name
```
fbergmann: the comment lies! ... it is comparing by 'id'

src/sbml/ListOf.h
```diff
@@ -604,6 +604,11 @@ class LIBSBML_EXTERN ListOf : public SBase
 
 
   /** @endcond */
+
+  /** @cond doxygenLibsbmlInternal */
+  void sort();
```
fbergmann, line 618: ideally there would be a docstring here, detailing that this function sorts the list based on the id of the Sbase elements.

src/sbml/test/TestListOf.c
```diff
@@ -215,6 +215,36 @@ START_TEST (test_ListOf_append)
...
+    ListOf_append(lo, sp);
+
+    lo->sort();
+    SBase_t* child = ListOf_get(lo, 0);
+    fail_unless(child->getId() == "a");
+    child = ListOf_get(lo, 1);
+    fail_unless(child->getId() == "n");
+    child = ListOf_get(lo, 2);
+    fail_unless(child->getId() == "z");
+
```
fbergmann, line 239: just wondering, what would be the expected outcome when sorting an L2 list of rules?

fbergmann (dismissed):
the sorting functionality will certainly be useful, just some comments came to mind that would be nice to sort.

src/sbml/test/TestListOf.c
luciansmith: This comment led me to adjust the sorting so it now sorts by, in order:
* the 'id' attribute
* the results of 'getId'
* the name
* the metaid.

So, if the IDs on the rules are set, it sorts by those.  If not, it sorts by the 'variable', since that's returned from a 'getId' call.

skeating (approved):

fbergmann (commented):
the changes were fine with me, i just had an issue with formatting, which i changed and committed, requesting review from you again @skeating

skeating (approved):

PR #61, opened 2020-10-29
Fix the allowed argument list for cauchy and laplace.

luciansmith:
The distrib-to-annotation converter assumes that the first element in the 'number of allowed arguments' list is 'the number of children allowed in the old annotations-only version'.  Since that's '2' for cauchy and laplace, the argument list for those functions is changed here to [2, 1, 4] from [1, 2, 4], so that the converter works.

This is tested in test_distrib_util, though I'm not sure if it's being run right now, since it was failing?

## Types of changes

- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ X] I have updated all documentation necessary.
- [ X] I have checked spelling in (new) comments.

## Testing


skeating (approved):
Sneaky :-)

PR #65, opened 2020-11-04
Issue 474 distrib flattening

luciansmith:
This addresses https://sourceforge.net/p/sbml/libsbml/474/?limit=25 and implements and tests distrib/comp flattening.

Something is still wonky with the flattening converter, because I could not get things to work as expected when 'distrib' was *not* enabled in the libsbml build:  it ends up coming through as an unknown package, but telling the converter to strip unflattenable packages apparently doesn't include unknown packages.  But that aside, everything else is now working and tested.

There were also two flattening tests that were failing that this commit fixes.  This means that travis isn't testing comp flattening, and may not be testing the packages at all?  That should also be fixed, but not here.

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/SBase.cpp
+++ b/src/sbml/SBase.cpp
@@ -236,10 +236,8 @@ SBase::prependStringToAllIdentifiers(const std::string& prefix)
   }
 
 
-  // for historical reasons some things like Rules will
-  // return an id but not set one
-  // so if we are going to bail on this we should have done
-  // anything else first
+  // Must use 'IdAttribute' functions since some elements like rules 
+  // return a different value for 'Id' functions alone.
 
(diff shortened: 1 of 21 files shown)


skeating (changes requested):
This does not pass all tests in test_comp_extensions.

Also, the issues you highlight would suggest more time/thought/investigation is needed

luciansmith:
The failing test is now fixed, though it was failing in development as well.

The issues I brought up should be addressed, but apply to *all* package flattening, not just distrib.  This PR brings distrib up to par with all the other packages, which I think is worth doing.

skeating (approved):

PR #75, opened 2020-11-13
Sonar units dir

skeating:
## Description

Fix the bugs listed by sonar cloud

## Types of changes

## Checklist:

## Testing


fbergmann (approved):
i really find the rule about the for loop for a non-issue, and would not have done that ... but yes, setting the ptr back to NULL is a good idea. Actually we maybe should have a macro like we have in COPASI pdelete, that will delete the given ptr, and always sets it to NULL afterwards. Would make the code look much cleaner, and can be easily checked later on.

PR #83, opened 2020-11-17
 - remove issues pointed out by sonar in sbml dir

fbergmann:
## Description
removed sonar issues from 'sbml' dir

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/Reaction.cpp
+++ b/src/sbml/Reaction.cpp
@@ -796,6 +796,9 @@ Reaction::unsetReversible ()
 int
 Reaction::addReactant (const SpeciesReference* sr)
 {
+  if (sr == NULL)
+    return LIBSBML_OPERATION_FAILED;
+
   int returnValue = checkCompatibility(static_cast<const SBase *>(sr));
   if (returnValue != LIBSBML_OPERATION_SUCCESS)
   {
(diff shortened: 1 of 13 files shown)


src/sbml/Reaction.cpp
```diff
@@ -1990,9 +2003,7 @@ Reaction::removeChildObject(const std::string& elementName, const std::string& i
 
   if (elementName == "kineticLaw")
   {
-    KineticLaw* t = getKineticLaw();
-    if (unsetKineticLaw() == LIBSBML_OPERATION_SUCCESS)
-      return t;
+    unsetKineticLaw(); // already deletes the kineticLaw, so nothing to do after
```
skeating, line 2006: surely there should be a return here

src/sbml/SpeciesReference.cpp
```diff
@@ -982,9 +982,7 @@ SpeciesReference::removeChildObject(const std::string& elementName, const std::s
 
   if (elementName == "stoichiometryMath")
   {
-    StoichiometryMath* t = getStoichiometryMath();
-    if (unsetStoichiometryMath() == LIBSBML_OPERATION_SUCCESS)
-      return t;
+    unsetStoichiometryMath();
```
skeating, line 985: return ?

skeating (changes requested):

src/sbml/Reaction.cpp
fbergmann: the return is at the end of the function ... so i didn't see a need to duplicate it here

src/sbml/SpeciesReference.cpp
fbergmann: the return is at the end of the function ... so i didn't see a need to duplicate it here, can change it though

skeating (approved):

PR #103, opened 2020-11-30
Fixing sonarcloud smells

JamieJQuinn:
## Description
Cleaned up most of the Sonar smells associated with UnitDefinition and SBMLDocument.

## Motivation and Context
These updates remove code smells as flagged by sonarcloud

## Types of changes

## Checklist:

## Testing

### DIFF
--- a/src/sbml/SBMLDocument.cpp
+++ b/src/sbml/SBMLDocument.cpp
@@ -286,7 +286,7 @@ SBMLDocument::SBMLDocument (const SBMLDocument& orig)
   
   if (orig.mModel != NULL) 
   {
-    mModel = orig.mModel->clone();
+    mModel = static_cast<Model*>( orig.mModel->clone() );
     mModel->setSBMLDocument(this);
   }
   
@@ -906,7 +906,7 @@ SBMLDocument::checkL1Compatibility (bool inConversion)
   unsigned int nerrors =  mInternalValidator->checkL1Compatibility();
   unsigned int unit_errors = 0;
 
-  if (not inConversion)
+  if (inConversion == false)
   {
     UnitConsistencyValidator unit_validator;
     unit_validator.init();
--- a/src/sbml/SBMLDocument.cpp
+++ b/src/sbml/SBMLDocument.cpp
@@ -1691,86 +1691,113 @@ SBMLDocument::readAttributes (const XMLAttributes& attributes,
   // The code for creating an extension IF below is almost equal to that in 
   // SBase::SBase(SBMLNamespaces*,typecode (int)).
   //
-  XMLNamespaces const *xmlns = getNamespaces();
+  XMLNamespaces *xmlns = getNamespaces();
   if (xmlns == NULL)
-    return;
-
-  int numxmlns= xmlns->getLength();
-  for (int i=0; i < numxmlns; i++)
   {
-    const std::string &uri = xmlns->getURI(i);
-    const SBMLExtension* sbmlext = SBMLExtensionRegistry::getInstance().getExtensionInternal(uri);
-
-    if (sbmlext && sbmlext->isEnabled())
+//     std::string err("SBase: xmlns is empty");
+//     XMLNamespaces* xmlns = sbmlns->getNamespaces();
+//     std::ostringstream oss;
+//     oss << "\nTypeCode " << typeCode << endl;
+//     if (xmlns)
+//     {
+//       XMLOutputStream xos(oss);
+//       xos << *xmlns;
+//     }
+//     err.append(oss.str());
+//     throw SBMLConstructorException(err);
+  }
+  else
+  {
+    int numxmlns= xmlns->getLength();
+    for (int i=0; i < numxmlns; i++)
     {
-      // if we are in l3v2 and there exists an l3v2 version for the package
-      // we wont accept the l3v1 version
-      if (sbmlext->getVersion(uri) < 2 && this->getVersion() > 1)
-      {
-        std::string dummyURI;
-        dummyURI.assign(uri);
-        size_t pos = dummyURI.find("level3");
-        if (pos != std::string::npos)
-        {
-          dummyURI.replace(pos, 15, "level3/version2");
-          if (sbmlext->getVersion(dummyURI) == 2)
-          {
-            ostringstream msg;
+      const std::string &uri = xmlns->getURI(i);
+      const SBMLExtension* sbmlext = SBMLExtensionRegistry::getInstance().getExtensionInternal(uri);
 
-            msg << "Package '" << xmlns->getPrefix(i) <<
-              "' has a L3V2V1 specification which must be used in an L3V2 document.";
-            logError(InvalidPackageLevelVersion, mLevel, mVersion, msg.str());
-            return;
+      if (sbmlext && sbmlext->isEnabled())
+      {
+        // if we are in l3v2 and there exists an l3v2 version for the package
+        // we wont accept the l3v1 version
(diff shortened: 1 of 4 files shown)


src/sbml/SBMLDocument.cpp
```diff
@@ -921,7 +906,7 @@ SBMLDocument::checkL1Compatibility (bool inConversion)
   unsigned int nerrors =  mInternalValidator->checkL1Compatibility();
   unsigned int unit_errors = 0;
 
-  if (inConversion == false)
+  if (not inConversion)
```
skeating: Snake in the grass :-) I think some python tried to sneak in

src/sbml/UnitDefinition.cpp
```diff
@@ -2475,13 +2395,13 @@ LIBSBML_EXTERN
 UnitDefinition_t*
 UnitDefinition_clone (const UnitDefinition_t *ud)
 {
-  return (ud != NULL) ? static_cast<UnitDefinition_t*>( ud->clone() ) : NULL;
```
skeating, line 2478: Okay these need to be left as essentially it is C code but being compiled by C++
a struct implies public inheritance by default ... class private inheritance by default
if that were to be enforced it could make a difference ... though i think we are always explicit with the inheritance we choose

so need to leave static casts

src/sbml/UnitDefinition.cpp
```diff
@@ -2669,23 +2589,23 @@ void
 UnitDefinition_simplify(UnitDefinition_t * ud)
 {
   if (ud != NULL) 
-    UnitDefinition::simplify(static_cast<UnitDefinition*>(ud));
```
skeating, line 2672: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2669,23 +2589,23 @@ void
...
+    UnitDefinition::simplify(ud);
 }
 
 LIBSBML_EXTERN
 void 
 UnitDefinition_reorder(UnitDefinition_t * ud)
 {
   if (ud != NULL) 
-    UnitDefinition::reorder(static_cast<UnitDefinition*>(ud));
```
skeating, line 2680: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2669,23 +2589,23 @@ void
...
+    UnitDefinition::reorder(ud);
 }
 
 LIBSBML_EXTERN
 UnitDefinition_t * 
 UnitDefinition_convertToSI(UnitDefinition_t * ud)
 {
   return (ud != NULL) ? 
-    UnitDefinition::convertToSI(static_cast<UnitDefinition*>(ud)) : NULL;
```
skeating, line 2688: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2706,8 +2625,7 @@ UnitDefinition_areEquivalent(UnitDefinition_t *ud1 , UnitDefinition_t * ud2)
 {
   if (ud1 != NULL && ud2 != NULL)
     return static_cast<int>(
-         UnitDefinition::areEquivalent(static_cast<UnitDefinition*>(ud1),
```
skeating, line 2709: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2694,8 +2614,7 @@ UnitDefinition_areIdentical(UnitDefinition_t * ud1, UnitDefinition_t * ud2)
 {
   if (ud1 != NULL && ud2 != NULL)
     return static_cast<int>
-         (UnitDefinition::areIdentical(static_cast<UnitDefinition*>(ud1),
```
skeating, line 2697: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2716,17 +2634,15 @@ LIBSBML_EXTERN
 UnitDefinition_t *  
 UnitDefinition_combine(UnitDefinition_t * ud1, UnitDefinition_t * ud2)
 {
-  return UnitDefinition::combine(static_cast<UnitDefinition*>(ud1),
```
skeating, line 2719: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2716,17 +2634,15 @@ LIBSBML_EXTERN
...
-                                       static_cast<UnitDefinition*>(ud2));
+  return UnitDefinition::combine(ud1, ud2);
 }
 
 
 LIBSBML_EXTERN
 UnitDefinition_t *  
 UnitDefinition_divide(UnitDefinition_t * ud1, UnitDefinition_t * ud2)
 {
-  return UnitDefinition::divide(static_cast<UnitDefinition*>(ud1),
```
skeating, line 2728: leave - as above

src/sbml/UnitDefinition.cpp
```diff
@@ -2735,8 +2651,7 @@ char *
 UnitDefinition_printUnits(UnitDefinition_t * ud, int compact)
 {
   return (ud != NULL) ? 
-    safe_strdup(UnitDefinition::printUnits(static_cast<UnitDefinition*>(ud)
```
skeating, line 2738: leave - as above

skeating (changes requested):

src/sbml/UnitDefinition.cpp
JamieJQuinn: I feel like I might be missing something here; I'm not sure how the inheritance policy changes how this pointer is handled. Here's how I understand it:

If `ud` represents some derived class/struct of `UnitDefinition_t` then the virtual method `ud->clone()` climbs down the inheritance tree and returns a pointer to the derived class. If the `static_cast` isn't there then that pointer is implicitly cast to `UnitDefinition_t*` on return. Unless I've missed something, the function of the static_cast is simply to perform this cast explicitly (though redundantly). Whether `ud` is an instance of a class or a struct doesn't seem to make a different.

skeating: Agreed in this case it probably doesnt BUT other classes with multiple inheritance it might !

src/sbml/SBMLDocument.cpp
```diff
@@ -278,25 +278,19 @@ SBMLDocument::SBMLDocument (const SBMLDocument& orig)
...
+  SBMLDocument::setSBMLDocument(this);
   
   mInternalValidator->setDocument(this);
   mInternalValidator->setApplicableValidators(orig.getApplicableValidators());
   mInternalValidator->setConversionValidators(orig.getConversionValidators());
   
   if (orig.mModel != NULL) 
   {
-    mModel = static_cast<Model*>( orig.mModel->clone() );
+    mModel = orig.mModel->clone();
```
skeating: With model the static_cast is necessary because in the comp package a ModelDefinition derives from Model

src/sbml/UnitDefinition.cpp
JamieJQuinn: Even in the case of multiple inheritance, because the return type of `UnitDefinition_clone` is `UnitDefinition_t*`, the return value of `ud-clone()` will be implicitly cast to the correct type, without the need for a static_cast. I can certainly keep the static_casts in if you feel strongly about it but I'm certain they're unnecessary in these cases.

src/sbml/SBMLDocument.cpp
JamieJQuinn: The cast can be done implicitly in this case, since `mModel` is defined as a `Model*`.

skeating (approved):
thank you

PR #104, opened 2020-11-30
Changed to use time.time() instead of time.clock()

funasoul:
## Description

Changed to use time.time() instead of time.clock() in readSBML.py

## Motivation and Context

Since the time.clock() function has been removed in python 3.8, changed to use time.time() instead.
See https://docs.python.org/3.7/library/time.html?highlight=time%20clock#time.clock

## Types of changes

## Checklist:

## Testing


fbergmann (approved):
thank you so much!

PR #107, opened 2020-12-07
Add continuous integration

alessandrofelder:
## Description

This PR adds CI to LibSBML. CI is implemented using GitHub actions, as GitHub Actions provides a higher number (20) of parallel jobs on more powerful machines than Travis, for free for open-source projects like ours. The file driving this is `cmake.yml` located in `.github/workflows` (There can be more such "workflow" YAML files there in the future if we want).

The implementation installs dependencies for, builds, and tests all combinations of 
- {MacOS, Ubuntu 16.04, Windows 10} operating systems 
- {LibXML2, expat, Xerces} xml parsers 
- with and without namespaces
- with and without examples
- with and without strict includes
which means that `3^2 * 2^3 = 9 * 8 = 72` independent jobs (which take ~1:30 h to complete in total) are run at every `push`. The implementation is pretty naive for now, but will cover a large part of the use cases at least. The only clever thing that happens is that the windows dependencies and SWIG download are cached. ~~I added an alternative SWIG path to a CMake file for convenience, as the structure of the underbelly of the Github Actions machine is still a bit murky for me.~~ 

I would be grateful for suggestions about
- how, if it all, can/should we test the CI itself (beyond looking at the output and saying that yes, `ctest` has passed some reasonable-looking tests).
- where/how to document the CI functionality (I will create a file on the shared drive for now, as Sarah suggested)
- anything else you think is important.
- routes for future improvement (I am currently trying a few :) )

## Motivation and Context
The context is described above. The change is required to start addressing the lack of CI/CD in LibSBML. It partially solves the problem that there is currently only minimal, if any, CI/CD for LibSBML.

## Types of changes

## Checklist:

## Testing
- [x] This cannot be tested automatically. I've done some manual reading of the GitHub Actions logs for the various combinations, and the output seems reasonable to me. I expected `Cmake` to catch any missing dependencies or settings, and the `ctest` runs at the end displayed reasonably named tests passing. The GitHub Actions logs can be viewed [here](https://github.com/alessandrofelder/libsbml/actions/runs/406153877)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -365,6 +365,7 @@ if(WITH_SWIG)
     find_program(SWIG_EXECUTABLE
         NAMES swig
         PATHS
+	      ./swig/swigwin-3.0.12
               c:/swigwin-3.0.10
               c:/swigwin-3.0.9
               c:/swigwin-3.0.8
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -365,7 +365,6 @@ if(WITH_SWIG)
     find_program(SWIG_EXECUTABLE
         NAMES swig
         PATHS
-	      ./swig/swigwin-3.0.12
               c:/swigwin-3.0.10
               c:/swigwin-3.0.9
               c:/swigwin-3.0.8
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -22,10 +22,10 @@ jobs:
         with_examples: ['True', 'False']
         package_option: ['', '-DWITH_STABLE_PACKAGES=ON', '-DWITH_ALL_PACKAGES=ON']
     runs-on: ${{ matrix.platform }}
-    
+
     steps:
     - uses: actions/checkout@v2
-    
+
     ### configure the operating system ###
     - name: Cache Windows dependencies and SWIG
       # On Windows, the dependencies live inside the source folder, ie `.`.
@@ -38,10 +38,10 @@ jobs:
               ./dependencies
               ./swig
         key: ${{ runner.os }}-dependencies
-    
+
     - name: Download pre-built Windows dependencies and SWIG
       # Windows dependencies have to be in a subfolder called `dependencies`, directly under the git repository root.
-      # also gets SWIG, completing the set of dependencies needed for windows      
+      # also gets SWIG, completing the set of dependencies needed for windows
       if: matrix.platform == 'windows-latest' && steps.cache-win-dependencies.outputs.cache-hit != 'true'
       shell: bash
       run: |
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -365,6 +365,7 @@ if(WITH_SWIG)
     find_program(SWIG_EXECUTABLE
         NAMES swig
         PATHS
+	      ./swig/swigwin-3.0.12
```
fbergmann: for the CI it would even be good to just specify the parameter manually -DSWIG_EXECUTABLE=<path to swig> during the configure stage. Adding now another long list of possible locations based on ./swig/.. is maybe not the best idea

alessandrofelder: Indeed, good idea - thanks.

alessandrofelder: I've managed to add SWIG to the path on the Windows CI in a clean way now and reverted the change to `CMakeLists.txt`.

alessandrofelder:
Currently, the CI runs on every push, which is possibly too often (given it takes ~1 hour and 30 minutes to complete). Would it make sense to run on PR/push to `development` and `stable` only?

fbergmann:
@alessandrofelder yes, i think all changes are basically pushed against the development branch, with stable only being touched, when we make a new release (and that one comes out of development). If someone needs additional branches checked, then we can enable that later on. 

and yes, lets make the full run only happen when a pull request is made. But lets have a quick run with all packages, just one xml parser, and one binding language happen at push. Would probably be fine to have that run only on the fastest platform for the build.

fbergmann:
what is important, is that at least one entry in the matrix runs with a full build against all packages, so with the option:  `-DWITH_ALL_PACKAGES=ON` .. and another one with stable packages `-DWITH_STABLE_PACKAGES=ON`. As these are currently untested.

alessandrofelder:
So, as discussed just before the winter break, I've added runs with `DWITH_ALL_PACKAGES=ON` and `DWITH_STABLE_PACKAGES=ON` to the CI config file, multiplying the number of runs by 3, from 72 to 216. This now takes 7 h 30 min to complete. I've also ensured that this will run only on new PRs to `development` or `stable`. I will add a shorter job to run on every `push` very soon. This means I think we've implemented all that we wanted to go into this current PR, so it would be good for you to have another review of this, @fbergmann , when you find the time :)  

[Edit] On the GH team plan, which I believe SBMLTeam is on, we should have 60 instead of 20 concurrent jobs, which should cut down the 7 h 30 min. Thanks to @giordano for pointing this out :)

In related news, I've rebased this PR onto the latest `development` today, and, if I understand correctly the CI has found a problem for the examples in Windows. `examples/c++/translateL3Math.cpp` reports an error `C2079: 'act' uses undefined struct 'main::sigaction'`. The only difference between the two latest CI runs on `alessandrofelder:add-continuous-integration`, which you can see [here](https://github.com/alessandrofelder/libsbml/actions?query=branch%3Aadd-continuous-integration++) is only that the last 65 commits from `sbmlteam:development` have been incorporated into the code via a `git rebase`. The second-to-last one (nr 152) passes, and the last one (nr 160) fails.

[This SO post](https://stackoverflow.com/questions/32389905/sigaction-and-porting-linux-code-to-windows#32390101) maybe suggests that we can't use all of `sigaction` in Windows, and therefore we might need to rethink the changes introduced [here](https://github.com/sbmlteam/libsbml/commit/c08a42431edebd8bcc891d249c381853fe038636) to `translateL3Math.cpp`? Caveat: I don't understand `sigaction` at all!
Shall I open an issue for now?

JamieJQuinn:
All the `sigaction` nonsense is just to get around a few sonarcloud code smells and seems to be introducing more problems than it solves. I would recommend simply reverting the offending commit.

fbergmann:
Instead of using signals, we could just check whether the last line was empty, and stop if it is. That would make sonar happy, as there will not be a `while(true)`, ctrl+c will still work to interrupt the program. And all should be fine. I won't get to it today.

fbergmann:
@alessandrofelder , could you try rebasing your patch again, and see whether it works now as intended? @JamieJQuinn kindly fixed the issue you had already.

alessandrofelder:
Great, thanks @JamieJQuinn ! 

Rebasing the latest changes onto a branch with just one configuration/operating system suggests the Jamie's fix did solve the issue. [The Windows job of this branch](https://github.com/alessandrofelder/libsbml/runs/1662620455?check_suite_focus=true) was failing before the latest rebase, but it passes now. 

Have now rebased and pushed to the current branch as well ~~- now we wait for 7 h 30 min :D~~. This worked as intended, @fbergmann  - I would say we can merge now?

fbergmann (approved):

PR #108, opened 2020-12-08
Fix sonarcloud smells 2

JamieJQuinn:
## Description
Fixed a number of code smells as flagged by SonarCloud in
- `SpeciesReference.cpp`
- `SBMLConvert.cpp`
- `SBase.cpp`

## Motivation and Context
This improves the maintainability of the code

## Types of changes

## Checklist:

## Testing
- [x] Testing is done automatically and codecov shows test coverage

### DIFF
--- a/src/sbml/SBase.cpp
+++ b/src/sbml/SBase.cpp
@@ -607,77 +607,32 @@ SBase::loadPlugins(SBMLNamespaces *sbmlns)
   //
   const XMLNamespaces *xmlns = sbmlns->getNamespaces();
 
-  if (xmlns)
+  if (!xmlns) return;
+
+  int numxmlns= xmlns->getLength();
+  SBaseExtensionPoint extPoint(getPackageName(), getTypeCode(), getElementName());
+  SBaseExtensionPoint genericPoint("all", SBML_GENERIC_SBASE);
+
+  for (int i=0; i < numxmlns; i++)
   {
-    int numxmlns= xmlns->getLength();
-    SBaseExtensionPoint extPoint(getPackageName(), getTypeCode(), getElementName());
-    SBaseExtensionPoint genericPoint("all", SBML_GENERIC_SBASE);
+    const std::string &uri = xmlns->getURI(i);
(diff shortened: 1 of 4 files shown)


src/sbml/SBase.cpp
```diff
@@ -451,8 +451,8 @@ SBase::SBase(const SBase& orig)
   if(orig.mCVTerms != NULL)
   {
     this->mCVTerms  = new List();
-    unsigned int i,iMax = orig.mCVTerms->getSize();
-    for(i = 0; i < iMax; ++i)
+    unsigned int iMax = orig.mCVTerms->getSize();
+    for(int i = 0; i < iMax; ++i)
```
skeating: should be unsigned int
or did sonar cloud have a rationle for this ?

src/sbml/SBase.cpp
```diff
@@ -7288,20 +7053,17 @@ SBase::updateSBMLNamespace(const std::string& package, unsigned int level,
         uri = SBML_XMLNS_L2V4;
         break;
       case 5:
-      default:
```
skeating, line 7291: Can you remember why removing dafults here was good ?
No problem if you can't.

skeating (approved):
Couple of question but more for my own interest

src/sbml/SBase.cpp
JamieJQuinn: Nope, that's my mistake.

src/sbml/SBase.cpp
JamieJQuinn: My rationale here was that it doesn't make sense for the default case to *actually* mean case 5, so the containing code should go in case 5. Sonarcloud was complaining about the empty `case` anyway. However, sonarcloud actually doesn't like switch-case blocks without defaults  (as I've left this) but I don't really know what the default *should* be in this case (I would assume an error message?) so I left the default case out so it remains flagged in sonarcloud to be fixed in the future.

I realise this is a bit complicated so let me know if you think it should be reverted.

skeating: Yes that one should be reverted as the default do correspond to the default levels and versions of sbml.

PR #115, opened 2021-01-13
split workflow into two

alessandrofelder:
previous workflow is now extensive.yml
brief.yml runs just one of extensive.yml matrix entries for each OS.

## Types of changes

## Checklist:

## Testing


fbergmann:
since i don't know how long i should wait before i approving the patch (as right now 8 / 203 checks are done)

for the language bindings, the following languages should be tested: 

* java & .net (as they have individual code written there)
* one other, maybe python ... (python, ruby, R, perl ...  use the same mechanism for the swig wrapping, so its mostly identical code there)
* later if at some point we figure out how, it would be nice to have matlab tested / built as well

the more i think about it ... having 203 separate runs is just too much

alessandrofelder:
For now, the PR is marked as draft, so no need to approve yet :) I will describe what the PR does in detail when it's ready, but the short of it is that it will run 3 jobs on `push` that take max 25 mins to finish, which is good to have, I think.

> the more i think about it ... having 203 separate runs is just too much

I agree! Though I am interested by how much `ccache` decreases the total execution time from 7 h 30 mins :)

I will make an effort to decrease the number of jobs a priority.

PR #117, opened 2021-01-17
Make the package buildable in CMake subdirectory

exaexa:
## Description
This allows to bundle libsbml in larger projects using `add_subdirectory` or CMake "ExternalProject" tools.

## Motivation and Context
Build of l3v2 math (and possibly other things) depended on the fact that libSBML is built from the "top" CMake directory. This fixes it (at least for the default build I tested).

Review is certainly needed though -- mainly, I am not sure about the rest of the build system, which occasionally contains raw `CMAKE_SOURCE_DIR` variable. Some additional changes might be needed to build e.g. the examples and some SWIG wrappers cleanly in a subdirectory (although there aren't many good reasons to do that).

## Types of changes

## Checklist:
- [x] I have updated all documentation necessary.   (Note: basically no documentation updated)
- [x] I have checked spelling in (new) comments.   (Note: no, there are none)

## Testing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 
 
 
-set(LIBSBML_ROOT_SOURCE_DIR "${CMAKE_SOURCE_DIR}" CACHE PATH
+set(LIBSBML_ROOT_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}" CACHE PATH
     "Path to the libSBML root source directory")
 set(LIBSBML_ROOT_BINARY_DIR "${CMAKE_BINARY_DIR}" CACHE PATH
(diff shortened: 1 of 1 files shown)


fbergmann:
Thank you for your contribution. 

this should not be necessary anymore, as in commit: 

https://github.com/sbmlteam/libsbml/commit/3d4f279a67d9db27381be7f9054f208269c50edf

lucian already made the same change. (and the binary dir would need updating too). could you verify that this works for you?

exaexa:
Hello @fbergmann,
oh that is cool, I did not notice that (since I was testing against the current official release). I will hotpatch the 5.19 version locally until the commit gets released. This also solves the PR. :] 
Thanks!
-mk

PR #120, opened 2021-01-31
Quick CI run on push

alessandrofelder:
## Description
This PR adds a fast workflow, detailed in `brief.yml`, that will be triggered at every push. This is for developers to get feedback as quickly as possible at every push. Furthermore, this PR adds functionality for draft PRs to skip the extensive CI build.

### A note on build times
Builds take ~3 minutes for Ubuntu, ~7 minutes for MacOS, and about 25 minutes for Windows. The large time difference to Windows is because `ccache` is not compatible with `ClangCL` or `MSVC`, the only compilers I managed to get to build the project successfully on Windows. The smaller time difference between MacOS and Ubuntu seems to be largely due to the longer amount of time it takes to download and install the dependencies with `brew`, which might be worth optimising in the future (e.g. using a GHA cache).

### A note on CI not passing
Checks for this CI were run twice ([first run](https://github.com/sbmlteam/libsbml/actions/runs/525793421), [second run](https://github.com/sbmlteam/libsbml/actions/runs/525793421)). One job (out of 216) failed during each of these runs, and it wasn't the same one. The second failure at least was because of some connection timing out. After discussion with @skeating this morning, we think this should be safe to ignore, but maybe still worth noting here.

## Motivation and Context
Currently, there is just one CI workflow on the development branch, which takes a lot of time, and is only run on PR (including draft PRs) to development. The motivation behind this PR is to provide developers with feedback as soon as possible after pushing, as well as to avoid draft PR triggering long builds.

## Types of changes

## Checklist:

- [x] I have updated all documentation necessary (CI docs are work-in-progress and currently in a document on our shared drive).

## Testing
- [x] This cannot be tested automatically - (tests can be seen passing [here](https://github.com/alessandrofelder/libsbml/actions/runs/525417194), although this does not prove that the CI itself is set-up well. It does look reasonable from looking at the logs though.)


fbergmann (approved):
i agree, the timeout does not have anything to do with the files committed here. But this will likely happen every once in a while when we have so many runs!

PR #121, opened 2021-02-16
Add basic nightly builds

alessandrofelder:
## Description

This PR adds a completely new GitHub actions workflow file, that is aimed at taking care of nightly builds.
The workflow is scheduled to run at 5 AM UTC (but actually seems to start around 15 minutes later) every day.
For each of Windows, Mac and Ubuntu, it builds and run the tests, with stable packages, and with all packages.
It then packages the builds into binaries and provides them as compressed files (zip for Windows, zip and tar.gz for unix-based OSs) with appropriate names. An example of the binaries provided can be seen [here](https://github.com/alessandrofelder/libsbml/actions/runs/570612644).

All CI currently only has Java bindings. Providing users with bindings to other languages will be done in a separate PR, which will also add those bindings to the two other workflows we have.

I am not sure what the best way to test the artifacts is? I would appreciate reviewer comments on this.

## Motivation and Context

This is a step towards a simpler and more frequent provision of the latest builds of libsbml.
More language bindings should be provided soon.

## Types of changes

## Checklist:

- [x] I have updated all documentation necessary (CI/CD docs currently held in a separate document on google drive).

## Testing
- [x] This cannot be tested automatically. Not sure how to best test this. Could manually download artefacts and check that they work well in conjunction with software they are supposed to integrate well with?

### DIFF
--- a/.github/workflows/store-artefact.yml
+++ b/.github/workflows/store-artefact.yml
@@ -1,6 +1,8 @@
 name: CMake (create and store Windows artefact)
 
-on: [push]
+on:
+  schedule:
+    - cron:  '0 5 * * *'
 
 env:
   BUILD_TYPE: Release
--- a/.github/workflows/store-artefact.yml
+++ b/.github/workflows/store-artefact.yml
@@ -108,7 +108,6 @@ jobs:
       - name: Build
         working-directory: ${{runner.workspace}}/build
         shell: bash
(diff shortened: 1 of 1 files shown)


fbergmann (approved):
looks good so far, i'll approve it and comment on the files later, once a run has gone through. thanks!

PR #122, opened 2021-02-18
Add language bindings (C#, C++, Python, Java) to all workflows

alessandrofelder:
## Description
This PR adds language bindings for C#, C++, Python to our three CI workflows ("extensive", "brief", and "artifact").

While adding these bindings, we noticed that a parallel `cmake` build can cause problems [for the C# bindings](https://github.com/alessandrofelder/libsbml/runs/1919060847?check_suite_focus=true) (which weren't tested before). I've therefore removed the parallel build option for now. One future way to enable parallel builds in the future would be to use Ninja as the generator.

Another change is that it's now necessary to specify `-DPYTHON_USE_DYNAMIC_LOOKUP=ON` for the Unix-based systems. Otherwise, some linking attempt fails on MacOS when trying to create the Python bindings (which also wasn't tested before).

A final change is to add a verbose flag to runs of `ctest` so more info on test failures will be provided by the CI by default.

I expect binaries created by the nightly build and based on the set of changes proposed here to appear early tomorrow morning on [the `development` branch of my fork](https://github.com/alessandrofelder/libsbml/actions). The nightly build only runs on `development` so the change will affect the binaries on `sbmlteam:development` only after merging.

## Motivation and Context
The typical language bindings that we want to test and provide with LibSBML are Java/C#/C++ and Python, but currently our CI/CD system only tests and provides Java bindings. The remaining of these binding are added by this PR. 

## Types of changes

## Checklist:

## Testing
- [x] This cannot be tested automatically. The automated tests do pass, but it's not clear (to me) whether there's a good way to test the CI workflow itself automatically. It would be good to test that the resulting artifacts are sensible manually, and this PR might make it easier to achieve this, as the binaries should now have all the bindings we want to provide to users included in the artifact.

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -112,7 +112,7 @@ jobs:
       - name: Build
         working-directory: ${{runner.workspace}}/build
         shell: bash
-        run: cmake --build . --config $BUILD_TYPE --parallel 4
+        run: cmake --build . --config $BUILD_TYPE
 
       ### run tests ###
       - name: Test
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -17,7 +17,10 @@ jobs:
         strict: ["True"]
         with_examples: ["True"]
         package_option: ["-DWITH_ALL_PACKAGES=ON"]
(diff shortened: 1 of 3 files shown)


fbergmann (approved):
that does look good to me. thanks for adding this

PR #124, opened 2021-02-23
reduce number of build configurations on PR

alessandrofelder:
## Description
Removes unnecessary workflow configurations from the 216 in `extensive.yml`, the "long" CI run that is triggered when a PR is made.
The removed workflow configurations are:
- "without any packages". This is unnecessary because any flaws with this configuration will be caught if we compiled with stable or all packages too.

## Motivation and Context
Currently, when we get to a PR stage, the `extensive.yml` GitHub actions workflow is run, consisting of 216 different jobs that take a total of ~ to run. Many of these runs are superfluous, and this PR is an attempt to remove the superfluous jobs from extensive.yml.

## Types of changes

## Checklist:

## Testing
- [x] This cannot be tested automatically. This PR only removes some superfluous workflow runs.

### DIFF
--- a/.github/workflows/extensive.yml
+++ b/.github/workflows/extensive.yml
@@ -22,7 +22,7 @@ jobs:
         strict: ["True", "False"]
         with_examples: ["True", "False"]
         package_option:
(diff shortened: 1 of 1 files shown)


alessandrofelder:
Closing in favour of a more sophisticated approach.

PR #126, opened 2021-02-24
Reduce number of jobs via conditionals

alessandrofelder:
## Description
Removes unnecessary workflow configurations from the 216 in extensive.yml, the "long" CI run that is triggered when a PR is made.

This is done using the following strategy (for each OS):
* All runs are done with examples, and using strict includes, with no exceptions.
* The XML parsers Expat and Xerces are only tested for one run each,  with C++ namespaces, with all packages, and with no language bindings.
* There are 6 runs using LibXML2 as the XML parser, all with C#, Java and Python language bindings, based on all combinations of
   * no packages, stable packages, all packages
   * with and without C++ namespaces

This results in 24 jobs in total, 8 jobs per OS -  a reduction of almost an order of magnitude in the number of jobs, and reduces the runtime of the extensive build from ~4.5 hours to about 1 hour.

## Motivation and Context
Currently, when we get to a PR stage, the extensive.yml GitHub actions workflow is run, consisting of 216 different jobs that take a total of ~4.5 to run. Many of these runs are superfluous, and this PR is an attempt to remove the superfluous jobs from extensive.yml.

## Types of changes

## Checklist:
- [x] I have updated all documentation necessary. CI docs kept in a word doc for now.

## Testing
- [x] This cannot be tested automatically. This PR only removes some superfluous workflow runs.

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -7,7 +7,7 @@ env:
 
 jobs:
   build:
-    name: ${{ matrix.platform }}, Parser option ${{ matrix.xml_parser_option }}, with namespaces ${{ matrix.with_namespace}}, strict includes ${{ matrix.with_namespace }}, with examples ${{ matrix.with_examples}}, package option ${{ matrix.package_option}}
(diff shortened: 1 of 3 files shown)


fbergmann (approved):
great! look forward to the faster turn arounds. I'll be sure to check the nightly archive tomorrow then.

alessandrofelder:
Looks like the "long" run, i.e. the [24 runs that took 1 hour](https://github.com/sbmlteam/libsbml/actions/runs/598901924) detailed in `extensive.yml`, finished successfully. Should we merge, @fbergmann ?

PR #131, opened 2021-03-19
Series of minor improvements to CI (Windows/zipping/C++ Standard)

alessandrofelder:
## Description
This PR addresses a series of minor flaws to the CI system.
Some of them are only Windows-related
- Windows libraries are built with static runtime, and use the static dependencies.
- A large file that is not needed in the binary is removed from the artifact

Additionally:
- ~~the rapid build tests both C++98 and C++20, spanning the range of current C++ standards~~ This is trickier than expected and will be addressed in a separate PR.
- the extensive build and the artifacts are explicitly built with C++98
- The artifacts are not doubly compressed (once explicitly by us, once implicitly be the `upload-artefact` action) - we leave the compression to the `upload-artefact` action, which creates a zip.

Note that the removal of the double compression creates much larger artifacts. The reviewer may wish to comment if this is a problem.

## Motivation and Context

Minor improvements to CI/CD.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change): The Windows artifacts are now different as they are built with static runtime and static dependencies (the other OS builds are also slightly different as the artifacts are not as compressed, and we specify the C++ standard for some of the builds).

## Checklist:

- [x] I have updated all documentation necessary. CI docs kept in separate document for now.

## Testing
- [x] This cannot be tested automatically. Artifacts look reasonable to me, and tests pass. Definitely worth the reviewer checking the artifacts too though.

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -50,7 +50,7 @@ jobs:
           curl -L https://prdownloads.sourceforge.net/swig/swigwin-3.0.12.zip > swig.zip
           unzip -o swig.zip -d swig
 
-      - name: add SWIG to Path (Windows)
+      - name: setup Windows environment
         # this is separate from the SWIG download itself, because it needs to be added to the path also when SWIG is cached
         if: matrix.platform == 'windows-latest'
         shell: bash
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -29,18 +29,18 @@ jobs:
         # On Windows, the dependencies live inside the source folder, ie `.`.
         # For the CI, we put SWIG there too, for simplicity.
         if: matrix.platform == 'windows-latest'
-        id: cache-win-dependencies
+        id: cache-win-dependencies-static
         uses: actions/cache@v2
         with:
           path: |
             ./dependencies
             ./swig
-          key: ${{ runner.os }}-dependencies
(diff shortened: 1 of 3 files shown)


fbergmann (commented):
looks fine, I was a bit surprised, just seeing the 98 standard singled out. Is it using the latest one otherwise? Or should we explicitly set it too for one build task?

alessandrofelder:
My understanding after discussing with @skeating last week was that we'd want to build the artifacts and the extensive build for C++98.
But we can change easily that if needed? I am not sure which one is used by default, but I think it might be C++98 anyway. I wanted to make it explicit in the `cmake` command though for clarity when reading.

The idea is to test both C++98 and C++20 in the rapid build, as there it doesn't cost anything (currently there are 3 jobs, and running each of those with both standards would result in 6 jobs that GH actions parallelises). I am in the process of making the changes [needed for C++20 and Visual C++, but encountered a problem with the Python bindings](https://github.com/alessandrofelder/libsbml/runs/2177047641?check_suite_focus=true) that I haven't managed to solve yet - so I've left that for the next PR....

fbergmann:
The problem there is that in c++17 there is no std::unexpected_handler, so that is a new issue you found! it was deprecated since c++11 and removed now. But that beyond this pr.

fbergmann (approved):

PR #148, opened 2021-06-07
Add `manylinux2010` nightly build

alessandrofelder:
## Description
This PR 
* adds functionality to build and test on `manylinux2010` at every push
* adds building, testing, and uploading of binaries on `manylinux2010` to the nightly build

These `manylinux2010` builds run as separate jobs in the same workflow, which keeps everything bundled nicely together. This is particularly important so we have all the binaries in the same place.

## Motivation and Context
Building and testing on `manylinux2010` ensures libSBML maintains a high degree of backwards compatibility.
The `manylinux2010` nightly build should allow us to reach a wider range of linux users.

Note that `v1` versions of the `checkout` and `upload-artefact` action had to be used to work successfully with `manylinux2010`. An effort to go back even further, to `manylinux1`, was abandoned due to compatibility issues with the `checkout` and `upload-artefact` action dependency `node.js`. It might be possible to provide `manylinux1` binaries with Github Actions if we used docker manually, but this will be left aside for now.

## Types of changes

- [ ] Change in documentation

## Checklist:

- [ ] I have updated all documentation necessary. Documentation of the CI is[ in preparation](https://github.com/alessandrofelder/libsbml/tree/add-initial-CI-docs).

## Testing
- [x] This cannot be tested automatically. Tests pass, otherwise the workflow would fail, but the tests are of course not testing the CI itself. It would be appreciated if the reviewer also could double-check the nightly build artefact created [here](https://github.com/alessandrofelder/libsbml/actions/runs/924669657) as a manual integration test. The two commits added after this was built won't change the contents of the artefact (they avoid `upload-artefact` specific warnings and ensure future `manylinux2010` nightly builds are not built on forks of `libSBML`.)


fbergmann (approved):
looks good to me thank you

PR #152, opened 2021-07-01
Add R builds to CI system

alessandrofelder:
## Description
This PR adds CI/CD related to R. More specifically, this PR:
- Full R binary packages are created for all OS at PR and Nightly Build. For unix-based systems, these runs include creating a source package first, and then compiling it into an R binary package with `R CMD INSTALL`.
- R binaries are built directly on push for ubuntu as part of the libSBML build.
- obsolete, macOS-specific linking options for R binaries are removed from `Makevars.in`.

## Motivation and Context
Solves the problem of needing to provide automatically nightly builds for libSBML R. 

## Types of changes

- [ ] Change in documentation

## Checklist:

- [ ] I have updated all documentation necessary. Full CI docs to follow in separate PR soon.

## Testing
- [x] This cannot be tested automatically. Tests pass (but this doesn't mean the CI setup is sensible). Contents of the created artefacts looks sensible to @alessandrofelder but the reviewer is encouraged to double-check. The artefacts created as part of [this run](https://github.com/alessandrofelder/libsbml/actions/runs/989682756) should match exactly what we'll get in the nightly build when this is merged.

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -22,7 +22,10 @@ jobs:
         with_examples: ["True"]
         package_option: ["-DWITH_ALL_PACKAGES=ON"]
         language_bindings:
-          ["-DWITH_JAVA=True -DWITH_CSHARP=True -DWITH_PYTHON=True"]
+          [
+            "-DWITH_JAVA=True -DWITH_CSHARP=True -DWITH_PYTHON=True
+            -DWITH_R=True",
+          ]
         cpp_standard: [98, 20]
     runs-on: ${{ matrix.platform }}
 
@@ -73,6 +76,9 @@ jobs:
         shell: bash
         run: |
           echo RUNTIME_LINKING_OPTION="-DWITH_STATIC_RUNTIME=ON" >> "${GITHUB_ENV}"
+          ./dev/utilities/expdef/expdef64.exe -dRlib.def -l R.dll
+          echo R_PLATFORM_SPECIFIC_OPTIONS="-DR_LIB=${GITHUB_WORKSPACE}\Rlib.lib" >> "${GITHUB_ENV}"
+          echo R_INCLUDE_PATH="/c/Program Files/R/R-4.1.0/include" >> "${GITHUB_ENV}"
 
       - name: Install Ubuntu dependencies
         # ubuntu already has SWIG and libxml2 by default
@@ -87,8 +93,15 @@ jobs:
         if: matrix.platform == 'macos-latest'
         shell: bash
         run: |
-          brew install check swig ccache
+          brew install check swig ccache gcc
           echo PYTHON_LINKING_OPTION="-DPYTHON_USE_DYNAMIC_LOOKUP=ON" >> "${GITHUB_ENV}"
+
+      - name: Unix R options
+        if: matrix.platform != 'windows-latest'
+        shell: bash
+        run:
+          echo R_PLATFORM_SPECIFIC_OPTIONS="-DWITH_CREATE_R_SOURCE=ON
+          -DWITH_SKIP_R_BINARY=ON" >> "${GITHUB_ENV}"
 
       ### setup ccache, not on Windows ###
       - name: Prepare ccache timestamp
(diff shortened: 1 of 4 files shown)


.github/workflows/brief.yml
```diff
@@ -81,6 +81,7 @@ jobs:
         run: |
           sudo apt-get install -y check ccache
           echo PYTHON_LINKING_OPTION="-DPYTHON_USE_DYNAMIC_LOOKUP=ON" >> "${GITHUB_ENV}"
+          echo R_BINDINGS="-DWITH_R=True"
```
fbergmann: i think you mean to pipe this into the environment too, right?

.github/workflows/extensive.yml
```diff
@@ -122,6 +122,9 @@ jobs:
         run: |
           echo $GITHUB_WORKSPACE"/swig/swigwin-3.0.12/" >> $GITHUB_PATH
           echo RUNTIME_LINKING_OPTION="-DWITH_STATIC_RUNTIME=ON" >> "${GITHUB_ENV}"
+          ./dev/utilities/expdef/expdef64.exe -dRlib.def -l R.dll
```
fbergmann, line 125: I'm surprised this found the R.dll, i would have expected to need the full path to the R runtime.

.github/workflows/store-artefact.yml
```diff
@@ -25,7 +25,10 @@ jobs:
         with_examples: ["True"]
         package_option: ["-DWITH_ALL_PACKAGES=ON", "-DWITH_STABLE_PACKAGES=ON"]
         language_bindings:
-          ["-DWITH_JAVA=True -DWITH_CSHARP=True -DWITH_PYTHON=True"]
+          [
+            "-DWITH_JAVA=True -DWITH_CSHARP=True -DWITH_PYTHON=True
+            -DWITH_R=True",
```
fbergmann, line 30: does the newline within the string cause an issue?

fbergmann:
@alessandrofelder  since you explicitly mention checking the created binaries, we still have the problem with the paths exported in the `lib/cmake/*release.cmake` files, they do absolute links to the libraries, like: 

    "D:/a/libsbml/libsbml/dependencies/lib/libbz2.lib;D:/a/libsbml/libsbml/dependencies/lib/zdll.lib;D:/a/libsbml/libsbml/dependencies/lib/libxml2.lib;WS2_32.lib;D:/a/libsbml/libsbml/dependencies/lib/libiconv.lib;D:/a/libsbml/libsbml/dependencies/lib/zdll.lib"

ideally we'd find a way to replace those absolute paths by `${_IMPORT_PREFIX}`. But that is not holding back this PR.

.github/workflows/brief.yml
alessandrofelder: Yep, thanks! Fixed.

.github/workflows/store-artefact.yml
alessandrofelder: This is the VS Studio Code autoformatting tool's fault. I am planning a formatting-only PR after this, but this slipped in now somehow.

It doesn't cause an issue and translates correctly in the run linked in the PR description to 
image

.github/workflows/extensive.yml
alessandrofelder: [This quick test](https://github.com/alessandrofelder/libsbml/runs/2964275078?check_suite_focus=true) suggests that the DLL is on the PATH of the GHA Windows runner: `which R.dll` evaluates correctly to `/c/Program Files/R/R-4.1.0/bin/x64/R.dll`.

alessandrofelder:
@fbergmann Is the lack of `${_IMPORT_PREFIX}` issue specific to Windows? Or does it the value of `IMPORTED_LINK_INTERFACE_LIBRARIES_RELEASE` in the `lib/cmake/*release.cmake` files that needs to be changed for all OS?
Just so I create an issue for this (Those paths seem to be absolute for all OS, but for the unix-based systems they at least point to what looks like the default location... whereas the prebuilt win dependencies live in the source repo...)

alessandrofelder:
Think this is ready to be merged when the tests have passed - I've addressed the review comments now.

fbergmann (approved):
thank you

PR #160, opened 2021-08-17
Implement functionality to infer reaction system from ODEs

alessandrofelder:
## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


src/sbml/math/ASTNode.cpp
```diff
@@ -3500,6 +3506,834 @@ void ASTNode::clearPlugins()
...
+      break;
+    case AST_TIMES:
+      derivative = derivativeTimes(variable);
+      break;
+    case AST_DIVIDE:
+      derivative = derivativeDivide(variable);
+      break;
+
+    default:
+      break;
```
fbergmann, line 4077: at this point its vital, that the derivative is initialized to a non-NULL value, as currently it will just crash without. Additionally, i feel the AST_POWER nodes have to be treated as well.

fbergmann: locally i avoid the crash using just `derivative = zero->deepCopy();` for now.

fbergmann:
please also add the registration of the converter as in: 

[converter-registry.diff.txt](https://github.com/sbmlteam/libsbml/files/7251324/converter-registry.diff.txt)

src/sbml/conversion/SBMLRateRuleConverter.cpp
```diff
@@ -0,0 +1,1090 @@
...
+  }
+}
+void
+SBMLRateRuleConverter::createReactions()
+{
+  unsigned int i = 0;
+  for (setCoeffIt it = mCoefficients.begin(); it != mCoefficients.end(); ++it)
+  {
+    Reaction *r = mDocument->getModel()->createReaction();
+    r->setReversible(false);
```
fbergmann, line 1038: please give the newly created reaction an id. any id will do. Leaving it empty will cause invalid models to be created. Some basic heuristic would be to use `J` + number of reactions in the model, or having that counter jut increase in case such an entity already exists in the model.

src/sbml/conversion/SBMLRateRuleConverter.cpp
```diff
@@ -0,0 +1,1090 @@
...
+  dealWithSpecies();
+  createReactions();
+  removeRules();
+}
+
+// helper function to determine whether a node is the "+" in a -x+y expression, where x,y are variable species.
+bool SBMLRateRuleConverter::isMinusXPlusY(ASTNode* node, Model* model)
+{
+    //if node is not binary, it's not -x+y
+    if (!node->getNumChildren() == 2)
```
fbergmann, line 898: this should probably be `node->getNumChildren() != 2`

src/sbml/conversion/SBMLRateRuleConverter.cpp
```diff
@@ -0,0 +1,1090 @@
...
+      }
+      if (mProducts[i][j] > 0)
+      {
+        stoichiometry = mProducts[i][j];
+        SpeciesReference *sr = r->createProduct();
+        sr->setSpecies(mODEs[j].first);
+        sr->setStoichiometry(stoichiometry);
+        sr->setConstant(true);
+      }
+      if (mModifiers[i][j] > 0)
```
fbergmann, line 1059: currently this leads to many modifiers being created if a symbol is used multiple times), please change to: 

`      if (mModifiers[i][j] > 0 && r->getModifier(mODEs[j].first) == NULL)`

src/sbml/conversion/SBMLRateRuleConverter.cpp
alessandrofelder: fixed.

src/sbml/conversion/SBMLRateRuleConverter.cpp
alessandrofelder: EDIT: Never mind - I understand now - sorry!

@fbergmann could you expand on this a bit for my understanding?
It naively seems like a weird check to add if in the subsequent lines we do
```c++
ModifierSpeciesReference *sr = r->createModifier();
sr->setSpecies(mODEs[j].first);
```
Why would we add a NULL species reference? or does `r->createModifier` change `mODEs`?

fbergmann: @alessandrofelder the idea is that `r->getModifier(mODEs[j].first)` would return `NULL` if it was not yet added as modifier. Without this, if a variable appears several times as modifier in `mODEs`, it would appear several times as modifier. When once is enough.

src/sbml/conversion/SBMLRateRuleConverter.cpp
alessandrofelder: Sure! Why `J`? What does `J` stand for here?

fbergmann: I'm fine with anything you are happy with, i just need a reaction id and any would do. `R` for reaction would work fine too ... in previous software we used `J`. I guess it refers to flux.

PR #162, opened 2021-08-24
Update Windows R Header version in CI

### LINKED ISSUE #161: Nightly build failure due to R binding issue
The last nightly build failed with: 

<https://github.com/sbmlteam/libsbml/actions/runs/1149525633>

looking into it it is caused by the R headers not being found, i assume this is because the build environment will have been upgraded to a higher version of R. So the action will need to be updated.

alessandrofelder:
## Description
Updates the hardcoded path to the windows R headers in the CI YAML files (the R version of the GitHub actions Windows VM changed last week).

I haven't found an easy way to prevent this failure happening when future R updates to the underlying OS configuration are made.
It may even help to leave this hard-coded so we know what R version we're using and notice when it changes (maybe?).

## Motivation and Context

Fixes #161 

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing
- [x] This cannot be tested automatically. I've tested that the `store-artefact.yml` workflow [completes successfully on push ](https://github.com/alessandrofelder/libsbml/runs/3410139674)(and then reverted the necessary change to achieve this test).


fbergmann (approved):
thanks

PR #167, opened 2021-09-22
 ubuntu 16.04 was removed

fbergmann:
## Description
nightly builds fail because the 16.04 ubuntu vm was removed, upgrading to 18.04

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [windows-latest, macos-latest, ubuntu-18.04]
-        xml_parser_option: ["-DWITH_LIBXML"]
+        xml_parser_option: ["-DWITH_EXPAT"]
         with_namespace: ["True"]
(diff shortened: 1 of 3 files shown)


alessandrofelder:
Should the CI docs be updated accordingly? Should we document (possibly somewhere else?) that certain combinations of Ubuntu version/XML parser/C++-standard are incompatible?

fbergmann:
> Should the CI docs be updated accordingly? Should we document (possibly somewhere else?) that certain combinations of Ubuntu version/XML parser/C++-standard are incompatible?

welcome back from vacation! yes, the documentation should mention that we now use the 18.04 container. I dont think the c++ standard of the individual libraries needs to be mentioned. That would just be another thing we'd have to update over and over again. Nothing would have been wrong with the CI build if we had not explicitly specified the c++ standard on those new machines. But as always the more documentation, the better, i just wanted these builds to work again.

skeating (approved):

PR #168, opened 2021-09-28
Fixes leak.

luciansmith:
If a LineEnding has a child BoundingBox, the BoundingBox that was created at the beginning of the function is leaked in favor of the new one.  This fix doesn't create a blank bounding box unless one doesn't exist.

### DIFF
--- a/src/fbc-package.cmake
+++ b/src/fbc-package.cmake
@@ -38,7 +38,7 @@ include(${LIBSBML_ROOT_SOURCE_DIR}/fbc-package.cmake)
 #build up sources
 set(FBC_SOURCES)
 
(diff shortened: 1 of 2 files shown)


luciansmith:
Not sure how to test this in libsbml (do we do memory checks anywhere now?), but I can say that I've tested the fix in roadrunner where I found the leak originally, and it worked there.

luciansmith:
I assume the failures are timeouts or the like, but I can't find where to tell it to re-run the failing tests.

skeating (changes requested):
You'll need to merge the ubuntu fix that is now in development branch

skeating (approved):

PR #169, opened 2021-10-01
Fixes the getErrorWithSeverity function.

luciansmith:
When the error in question is an XMLError and not an SBMLError, the old dynamic_cast fails and returns NULL.  This is I think because the run-time tests that dynamic_cast try to run cannot run because XMLError has no virtual functions, and the test cannot be run, and dynamic_cast returns NULL.  Changing to a static_cast works instead, because it has no checks.

### DIFF
--- a/src/sbml/SBMLError.h
+++ b/src/sbml/SBMLError.h
@@ -1253,6 +1253,10 @@ protected:
 
   void adjustErrorId(unsigned int offset);
 
+  virtual void setMessage(const std::string& message);
+
+  friend class SBMLErrorLog;
+
   /** @endcond **/
 };
 
--- a/src/sbml/SBMLError.h
+++ b/src/sbml/SBMLError.h
@@ -1253,10 +1253,6 @@ protected:
 
   void adjustErrorId(unsigned int offset);
 
-  virtual void setMessage(const std::string& message);
-
-  friend class SBMLErrorLog;
-
   /** @endcond **/
 };
 
--- a/src/sbml/SBMLErrorLog.cpp
+++ b/src/sbml/SBMLErrorLog.cpp
@@ -173,6 +173,27 @@ SBMLErrorLog::add (const SBMLError& error)
 
 
 /*
+ * Adds the given XMLError to the log.
+ *
+ * @param error SBMLError, the error to be logged.
+ */
+void
+SBMLErrorLog::add(const XMLError& error)
+{
+    if (error.getSeverity() == LIBSBML_SEV_NOT_APPLICABLE)
+        return;
+    const SBMLError* serror = dynamic_cast<const SBMLError*>(&error);
+    if (serror == NULL)
+    {
+        SBMLError sbmlerr(error.getErrorId(), 0, 0, "", error.getLine(), error.getColumn(), error.getSeverity(), error.getCategory());
+        sbmlerr.setMessage(error.getMessage());
+        return XMLErrorLog::add(sbmlerr);
+    }
+    XMLErrorLog::add(error);
+}
+
+
+/*
  * Logs (copies) the SBMLErrors in the given SBMLError list to this
  * SBMLErrorLog.
  *
--- a/src/sbml/SBMLErrorLog.cpp
+++ b/src/sbml/SBMLErrorLog.cpp
@@ -364,7 +364,7 @@ SBMLErrorLog::getNumFailsWithSeverity(unsigned int severity)
 const SBMLError*
 SBMLErrorLog::getError (unsigned int n) const
 {
-  return static_cast<const SBMLError*>(XMLErrorLog::getError(n));
+  return dynamic_cast<const SBMLError*>(XMLErrorLog::getError(n));
 }
 
 /*
@@ -385,7 +385,7 @@ SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
   {
     if (matcher(*it))
     {
-      if (count == n) return static_cast<const SBMLError*>(*it);
+      if (count == n) return dynamic_cast<const SBMLError*>(*it);
       ++count;
     }
   }
--- a/src/sbml/SBMLErrorLog.cpp
+++ b/src/sbml/SBMLErrorLog.cpp
@@ -385,7 +385,7 @@ SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
   {
     if (matcher(*it))
     {
-      if (count == n) return dynamic_cast<const SBMLError*>(*it);
+      if (count == n) return static_cast<const SBMLError*>(*it);
       ++count;
     }
   }
--- a/src/sbml/test/TestSBMLError.cpp
+++ b/src/sbml/test/TestSBMLError.cpp
@@ -242,6 +242,10 @@ START_TEST(test_SBMLErrorLog_getErrorWithSeverity_XMLError)
     fail_unless(error != NULL);
     fail_unless(error->getErrorId() == XMLContentEmpty);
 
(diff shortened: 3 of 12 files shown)


skeating (changes requested):
This looks like its breaking an existing test in the sbml dir
sorry dont have time to look 
hopefully be freer by end of week

src/sbml/SBMLErrorLog.cpp
```diff
@@ -385,7 +385,7 @@ SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
   {
     if (matcher(*it))
     {
-      if (count == n) return dynamic_cast<const SBMLError*>(*it);
+      if (count == n) return static_cast<const SBMLError*>(*it);
```
fbergmann: this cannot work, since the iterator could be pointing to an error of type `XMLError`, which is the base class of SBML error. So the dynamic cast was there to return a NULL error in that case. The static cast would just force the error object into an SBML error, and create a bad object. 

Thus the test failures: 

```

2021-10-01T19:34:00.9150270Z 35:  Generic Attribute Functions
2021-10-01T19:34:00.9655721Z 35: 99%: Checks: 1737, Failures: 1, Errors: 0
2021-10-01T19:34:00.9657744Z 35: D:\a\libsbml\libsbml\src\sbml\test\TestSBMLError.cpp:243:F:SBMLError:test_SBMLErrorLog_getErrorWithSeverity_XMLError:0: Assertion 'error->getErrorId() == XMLContentEmpty' failed
2021-10-01T19:34:00.9695739Z  35/184 Test  #35: test_sbml_sbml_run ............................***Failed    2.04 sec
2021-10-01T19:34:00.9700201Z test 36
```

src/sbml/test/TestSBMLError.cpp
```diff
@@ -173,6 +175,76 @@ START_TEST(test_SBMLErrorLog_removeAll)
...
+}
+END_TEST
+
+START_TEST(test_SBMLErrorLog_getErrorWithSeverity_XMLError)
+{
+    //NOTE:  This test failed originally because an XMLError cannot be dynamically cast to an SBMLError; it must be statically cast.  I believe this is because the XMLError has no functions for the dynamic cast checker to test, so it returns NULL.
+    SBMLDocument* doc = readSBMLFromString("Not actually SBML.");
+
+    const SBMLError* error = doc->getErrorWithSeverity(0, LIBSBML_SEV_ERROR);
+    fail_unless(error != NULL);
```
fbergmann, line 242: I dont think this should pass.

luciansmith:
In that case, 'getError' should fail, then, since it uses static_cast.  But nobody's complained about that, and the static_cast works for me on Windows, so I'm not sure what the correct solution is.

luciansmith:
OK, tried the upconverting thing.  Test work (again) on Windows; hopefully they'll work elsewhere too.

src/sbml/xml/XMLErrorLog.cpp
```diff
@@ -126,6 +126,14 @@ XMLErrorLog::add (const XMLError& error)
 {
   if (mOverriddenSeverity == LIBSBML_OVERRIDE_DONT_LOG) return;
 
+  const SBMLError* serror = dynamic_cast<const SBMLError*>(&error);
+  if (serror == NULL)
+  {
+      SBMLError sbmlerr(error.getErrorId(), 0, 0, "", error.getLine(), error.getColumn(), error.getSeverity(), error.getCategory());
```
fbergmann: @luciansmith I didnt' think you would add the upconversion here. This is potentially problematic, as sarah was working on liblx, which essentially removes the XML layer from libSBML (and all references to SBML), so i was expecting you to add this functionality at the 'getError' level of the sbml document, for the cases where you get a non-sbml error returned. but @skeating  will have to clearify.

luciansmith: If 'getError' returned an object, it would be easy to upconvert there, but since it returns a pointer, you need to save it somewhere.  I suppose it could delete and re-add the object at that point?

Another option might be to make 'add' virtual, and override it with SBMLError?  The problem I was running into was that I couldn't access member variables that way.

luciansmith: Done!  Moved the upcast to SBMLErrorLog's 'add', by virtue of making XMLErrorLog's 'add' virtual, and adding a 'setMessage' function to SBMLError (protected) and setting the SBMLErrorLog as a friend.

src/sbml/SBMLErrorLog.cpp
```diff
@@ -172,6 +172,27 @@ SBMLErrorLog::add (const SBMLError& error)
...
+SBMLErrorLog::add(const XMLError& error)
+{
+    if (error.getSeverity() == LIBSBML_SEV_NOT_APPLICABLE)
+        return;
+    const SBMLError* serror = dynamic_cast<const SBMLError*>(&error);
+    if (serror == NULL)
+    {
+        SBMLError sbmlerr(error.getErrorId(), 0, 0, "", error.getLine(), error.getColumn(), error.getSeverity(), error.getCategory());
+        sbmlerr.setMessage(error.getMessage());
+        return XMLErrorLog::add(sbmlerr);
```
fbergmann, line 190: that return statement looks odd ... when looking at the void method. Might be better to have it after the add statement,

src/sbml/SBMLError.h
```diff
@@ -1253,6 +1253,10 @@ class LIBSBML_EXTERN SBMLError : public XMLError
 
   void adjustErrorId(unsigned int offset);
 
+  virtual void setMessage(const std::string& message);
+
+  friend class SBMLErrorLog;
```
fbergmann: why would we need the friend statement, just because setMessage is private? I think i actually wouldn't mind a public `setMessage`, as then people could customize error messages more easily.

luciansmith: This follows the current design of XMLError, which is that it cannot be changed after construction.  This has always seemed like an odd decision to me, but it's baked into the design from the get go.  Here, we *have* to allow message editing, because you cannot construct the object with the entire message; the automatic message will always be added first, so this allows for that situation while making the smallest design change to XMLError.

I would be all for changing the design so that XMLError itself just made public 'setX' functions for everything, but that's a different issue.

src/sbml/SBMLErrorLog.cpp
luciansmith: I dunno; I feel like it's a pretty common idiom.  We're calling the base class version of the function and doing the same thing.  This also doesn't have to be changed in the future if we decide to (say) change the signature so it returns an error code on failure.

luciansmith: (It's obviously fine to expand it to two lines if anyone really wants it, if course.)

src/sbml/SBMLError.h
skeating: Actually that decision even predates me arrgg

Yes I'm happy to with allowing Errors to be edited after creation - this allows people flexibility.
So pubic setMessage/setSeverity

:-)

luciansmith: In that case, I'm inclined to go ahead and change XMLError itself--will that mess with anything on the libx side of things?

fbergmann: @luciansmith i think that is an improvement, that liblx will also have to mirror. Thanks for doing this.

fbergmann (approved):
looks all good to me

luciansmith:
@skeating : Does this work for you now?  You have a requested change.

skeating (approved):

PR #172, opened 2021-10-07
Fix spatial validation crash.

luciansmith:
When checking to see if spatial ids are unique, would crash if boundaryMax or boundaryMin objects were missing.  This adds a check to prevent this.


fbergmann (approved):
good catch

PR #185, opened 2021-12-07
Use string ref for get

luciansmith:
Simple speedup fix:  when asking for an element by id or metaid, use const string& instead of copying the id string (which doesn't change).

Noticed this when profiling libroadrunner.


luciansmith:
I have no idea where the other four commits came from; I just synced to develop and am slightly changing two files?  Git is weird.

PR #186, opened 2021-12-07
Const ref for get element

luciansmith:
Noticed when profiling libroadrunner that 'getElementBySId' was malloc'ing and free'ing a lot of strings; this turned out to be the culprit.

Did a regexp search for 

'::.*get.*string '

which found that function, the same for '..byMetaId', and several functions in the comp plugin that did the same thing.  Also typography fixes added for all 'string &uri' -> 'string& uri', since that's the form we use literally everywhere else, and made the regexp work.


luciansmith:
I cannot understand where the extra four commits are coming from!  It's super mysterious.  And three of them are 'merge development into development'!  What the heck.

skeating (approved):

PR #198, opened 2022-02-07
add `WITH_THREADSAFE_PARSER` option

### LINKED ISSUE #197: `SBML_parseL3FormulaWithSettings` thread safety
Currently `SBML_parseL3FormulaWithSettings` is not safe to call from multiple threads at the same time, as it uses & modifies the single global instance of `L3Parser`, so calling it from multiple threads simultaneously results in garbage data and/or segfaults.

One possible solution could be to add a mutex, so that only a single thread can mutate the L3Parser object at any time.
This would require c++11, but it could be optionally enabled or disabled at compile time with e.g. `-DWITH_THREADSAFE_PARSER=ON`

lkeegan:
## Description
- add CMake option `WITH_THREADSAFE_PARSER`, default to OFF
- if enabled
  - adds a mutex to allow `SBML_parseL3FormulaWithSettings` to be safely called from multiple threads
  - this requires c++11 or greater

## Motivation and Context
fixes #197 

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,6 +315,11 @@ option(WITH_XERCES   "Use the Xerces XML parser library."    OFF)
 # Use C++ namespace.
 option(WITH_CPP_NAMESPACE "Use a C++ namespace for libSBML."   OFF)
 
+option(WITH_THREADSAFE_PARSER "Use a threadsafe parser."   OFF)
+if(WITH_THREADSAFE_PARSER)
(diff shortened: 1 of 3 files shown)


fbergmann (approved):
thank  you for the fix. We will be discussing, whether we can increase the default  c++ standard to C++11, which seems appropriate. Until then the fix can remain as cmake option.

PR #205, opened 2022-03-17
Cmake find modules

fbergmann:
## Description
The cmake build gives several warnings currently, the exported target cmake files contain full path to libraries, which makes them not relocatable. 

This PR sorts these points, for a couple of libraries for now. 

## Motivation and Context
Make it easier to link against libSBML when importing CMAKE targets. 

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


fbergmann:
(the test failure of the one osx run was because of a timeout of the build machine)

fbergmann:
apparently the ubuntu 18.04 runners don't work right now either. but I've verified that this is not an issue. with the changes I made.

giordano:
> apparently the ubuntu 18.04 runners don't work right now either. but I've verified that this is not an issue. with the changes I made.

Running `sudo apt-get update`, to update the list of packages, before `sudo apt-get install` should help.  There has been a security update for `expat` a few days ago: http://changelogs.ubuntu.com/changelogs/pool/main/e/expat/expat_2.2.5-3ubuntu0.7/changelog

fbergmann:
that would increase the build time quite a bit, but could be done. However, this has nothing to do with this pr.

giordano:
> that would increase the build time quite a bit

Why?  `apt-get update` should take a few seconds.  To be clear, it doesn't update all the packages in the environment (that's `apt-get upgrade`), only the list of the available packages in the repositories, which is the issue here: the package manager is trying to install the wrong version of a package.

skeating (approved):

PR #210, opened 2022-04-05
Independent cmake modules

fbergmann:
## Description
The find scripts we introduced work really nice. However, every library consuming libSBML would have to use the same find scripts. What is more, since we hard coded `LIBSBML_DEPENDENCY_DIR`, every other consumer would now have to define this variable. 

This PR does 2 things: 

* the dependency dir variable is computed automatically from the PROJECT_NAME defined. So consumers can just define For example a `LIBSEDML_DEPENDENCY_DIR` variable for the LIBSEDML project. And the values will be read. 
* all find modules are installed into share/cmake/Modules

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


skeating (approved):
cool

PR #214, opened 2022-04-12
Fix lols

### LINKED ISSUE #213: libsbml & billion lols
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2003-1564

Also, we're going to be used as an example of 'old bugs still being used in modern libraries' so it probably behooves us to update.

fbergmann:
## Description
This PR aims to fix the issues we have with the laughing attacks. 

## Motivation and Context
 - fixes #213 
 - also requires some changes with updated to latest parsers. 

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


fbergmann:
the test failure on ubuntu is due to the expat version on ubuntu is too old, and the issue is still present there.

fbergmann:
the windows test fails, as the downloaded binaries were from the old dependency set, so updating the dependencies will be next. With the caveat, that the latest expat / libxml require a quite high version of visual studio.

PR #216, opened 2022-04-20
chore: clarify getPackageRequired's doc

### LINKED ISSUE #215: getPackageRequired ignores package name
## Issue description

The full model:
[SbmlWithComp.log](https://github.com/sbmlteam/libsbml/files/8519941/SbmlWithComp.log) (change the extension from `.log` to `.xml` to reproduce, github does not accept the latter for attachments)

The interesting bit:
```xml
<sbml xmlns:comp="http://www.sbml.org/sbml/level3/version1/comp/version1" xmlns="http://www.sbml.org/sbml/level3/version2/core" level="3" version="2" comp:required="true">
  <model id="case01355" name="case01355">
```

When loading the above SBML model, the `getPackageRequired` function invariably returns `0` when supplied `"comp"` as package parameter.
However [the doc for this function](https://sbml.org/software/libsbml/5.18.0/docs/formatted/c-api/class_s_b_m_l_document__t.html#abf9dde83144945cd14bead4717e8afdd) suggests that one can pass either the package name or the package URI:
```
int SBMLDocument_getPackageRequired	(	SBMLDocument_t * 	d,
const char * 	package 
)		
Returns the required attribute of the given package extension.

Parameters
d	the SBMLDocument_t structure to check.
package	the name or URI of the package extension.
Returns
1 (true) if the package is flagged as being required in this SBML document, 0 (false) otherwise.
```

Passing the full URI `"http://www.sbml.org/sbml/level3/version1/comp/version1"` works as expected, _i.e._ returns `0` when `comp:required="false"` and `1` when `comp:required="true"`.

eltix:
## Description

Clarified the doc for the `getPackageRequired` function.

## Motivation and Context

The behavior of the function depends on some configuration options which makes it confusing for the user. According to the current documentation, the function allows passing either the package name (or "nickname") or its URI. However, passing the package name only works if support for that particular package was activated when compiling the library.

Fixes #215 

## Types of changes

- [x] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/SBMLDocument.h
+++ b/src/sbml/SBMLDocument.h
@@ -1391,7 +1391,10 @@ public:
    * prefix will be written when writing elements defined in the given
    * package extension if @c true is given as second argument.
    *
-   * @param package the name or URI of the package extension.
+   * @param package the name or URI of the package extension. Passing a package
+   * name (or "nickname") is only supported if libSBML was compiled with support for
+   * that particular package, see the installation documentation for more details.
+   * Passing the package URI is supported regardless of the installation configuration.
    * @param flag boolean value to indicate whether to write a namespace
    * prefix.
    *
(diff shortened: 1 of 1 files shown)


fbergmann:
Thank you for this, we probably should change the other 13 places that used the same description as well.

eltix:
@fbergmann Do you want me to do it for the other 13 or do you have some smart factorization in mind?

fbergmann:
@eltix I would have just searched & replaced the remaining entries. Were you thinking of something else?

eltix:
@fbergmann Ok, I've done just that.

fbergmann (approved):
thank you so much!

PR #218, opened 2022-04-21
Fix #217: coordinateBoundary validation.

### LINKED ISSUE #217: Spatial validation rule 23606 fails to notice 'min' boundaries.
It complains about coordinateBoundary objects that reference minimums.

luciansmith:
Fix issue #217


fbergmann (approved):
good catch

PR #222, opened 2022-04-29
The const version of getElementBySId should be virtual.

luciansmith:
This makes it match everything else, and also means it gets properly overwritten when it should be.


fbergmann (approved):
works for me

PR #223, opened 2022-04-29
Spatial update

luciansmith:
Remove code that corresponds to the stuff we just removed from spatial, namely:

* quadrilateral meshes.
* 'difference' between sets (aka relative complements)
* The 'complementA' and 'complementB' attributes, which were only used for 'difference'.


fbergmann (approved):
that looks good to me. Thanks

PR #243, opened 2022-07-13
 #241:issues with fbc v3

### LINKED ISSUE #241: `fbc-v3`: Can't set xmlns on `KeyValuePairList` & Can't set key and value on `KeyValuePair`
Hi all,
we are currently in the process of implementing support for the fbc version 3 draft in `sbmlutils` and `cobrapy`.
I just tried creating the `KeyValuePairs` but have some issues.

The following code results in issues

```python
"""
Test bed for FBC version 3.
For latest SBML fbc v3 specification see
https://sourceforge.net/p/sbml/code/26312/tree/trunk/specifications/sbml-level-3/version-1/fbc/spec/main.pdf?format=raw

"""
import libsbml
from logging import getLogger
logger = getLogger(__name__)


def check(value: int, message: str) -> bool:
    """Check the libsbml return value and prints message if something happened.

    If 'value' is None, prints an error message constructed using
      'message' and then exits with status code 1. If 'value' is an integer,
      it assumes it is a libSBML return status code. If the code value is
      LIBSBML_OPERATION_SUCCESS, returns without further action; if it is not,
      prints an error message constructed using 'message' along with text from
      libSBML explaining the meaning of the code, and exits with status code 1.
    """
    valid = True
    if value is None:
        logger.error(f"Error: LibSBML returned a null value trying to <{message}>.")
        valid = False
    elif isinstance(value, int):
        if value != libsbml.LIBSBML_OPERATION_SUCCESS:
            logger.error(f"Error encountered trying to '{message}'.")
            logger.error(
                f"LibSBML returned error code {str(value)}: "
                f"{libsbml.OperationReturnValue_toString(value).strip()}"
            )
            valid = False

    return valid


# create document with new fbc version 3 namespace
sbmlns: libsbml.SBMLNamespaces = libsbml.SBMLNamespaces(3, 1)
sbmlns.addPkgNamespace("fbc", 3)
doc: libsbml.SBMLDocument = libsbml.SBMLDocument(sbmlns)
doc_fbc: libsbml.FbcSBMLDocumentPlugin = doc.getPlugin("fbc")
doc_fbc.setRequired(False)

model: libsbml.Model = doc.createModel()
model_fbc: libsbml.FbcModelPlugin = model.getPlugin("fbc")
print(model_fbc)
model_fbc.setStrict(True)

# Support for key value pairs exists
c: libsbml.Compartment = model.createCompartment()
c.setId("c1")
c.setSize(1.0)
c.setConstant(True)

s1: libsbml.Species = model.createSpecies()
s1.setId("s1")
s1.setCompartment("c1")
s1.setInitialConcentration(1.0)
s1.setConstant(True)
s1.setHasOnlySubstanceUnits(False)
s1.setBoundaryCondition(False)

s1_fbc: libsbml.FbcSpeciesPlugin = s1.getPlugin("fbc")

kvp_list: libsbml.ListOfKeyValuePairs = s1_fbc.getListOfKeyValuePairs()
kvp_list.setXmlns("http://sbml.org/fbc/keyvaluepair")
kvp: libsbml.KeyValuePair = s1_fbc.createKeyValuePair()
check(kvp.setKey("testdata"), "Set Key on KeyValuePair")
check(kvp.setValue("1.0"), "Set Value on KeyValuePair")
s1_fbc.addKeyValuePair(kvp)

sbml_str: str = libsbml.writeSBMLToString(doc)
print("-" * 80)
print(sbml_str)
print("-" * 80)
```

Results in problems setting the KeyValue Pairs? You can see that these are not written in the SBML below.
Also the xlmns is not set on the list, but `xmlns="http://www.sbml.org/sbml/level3/version1/fbc/version3"`. Perhaps this is the reason the setting of Key and Value fails ?

The specification states that 
> The xmlns is a mandatory component of the ListOfKeyValuePairs, is of the type uri and must have the value
`http://sbml.org/fbc/keyvaluepair`. 

But it seems I cannot set the xmlns. Also why is this not set automatically if the value has the required value `http://sbml.org/fbc/keyvaluepair`

```
<FbcModelPlugin>
Error encountered trying to 'Set Key on KeyValuePair'.
LibSBML returned error code -2: The attribute that is the subject of this operation is not valid for the combination of SBML Level and Version for the underlying object. This can happen because libSBML strives to offer a uniform API for all SBML Levels and Versions, but some object attributes and elements are not defined for all SBML Levels and Versions. Calling programs are expected to be aware of which object structures they are working with, but when errors of this kind occur, they are reported using this return value.
Error encountered trying to 'Set Value on KeyValuePair'.
LibSBML returned error code -2: The attribute that is the subject of this operation is not valid for the combination of SBML Level and Version for the underlying object. This can happen because libSBML strives to offer a uniform API for all SBML Levels and Versions, but some object attributes and elements are not defined for all SBML Levels and Versions. Calling programs are expected to be aware of which object structures they are working with, but when errors of this kind occur, they are reported using this return value.
```
and the following SBML
```
<sbml xmlns="http://www.sbml.org/sbml/level3/version1/core" xmlns:fbc="http://www.sbml.org/sbml/level3/version1/fbc/version3" level="3" version="1" fbc:required="false">
  <model fbc:strict="true">
    <listOfCompartments>
      <compartment id="c1" size="1" constant="true"/>
    </listOfCompartments>
    <listOfSpecies>
      <species id="s1" compartment="c1" initialConcentration="1" hasOnlySubstanceUnits="false" boundaryCondition="false" constant="true">
        <annotation>
          <listOfKeyValuePairs xmlns="http://www.sbml.org/sbml/level3/version1/fbc/version3">
            <keyValuePair/>
          </listOfKeyValuePairs>
        </annotation>
      </species>
    </listOfSpecies>
  </model>
</sbml>
```

Best Matthias

fbergmann:
## Description
in #241 matthias noticed a couple of inconsistencies. 

* need to ensure correct package version is passed along
* the xmlns on the list of key value pairs is wrong. 

## Motivation and Context
fixes #241

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


fbergmann:
i've confirmed that all issues are adressed,  and working (here a colab book: https://colab.research.google.com/drive/1WTkKmnk8dbJ1NoMedlc91gitDdc2-P8U?usp=sharing) and corresponding python wheels for it. https://github.com/sbmlteam/python-libsbml/actions/runs/2675478034

PR #276, opened 2022-10-27
#275: check that spatial dimensions match before running test

### LINKED ISSUE #275: spatial validation rule 1221350 is to stringent
@jcschaff noticed, that libsbml is flagging compartmentmappings unitSize as wrong when it should not. lisbm complaints that: 

```
 The attribute 'spatial:unitSize' on a <compartmentMapping> must have a value between 0 and 1, inclusive.
```

while the specification says: 

```
3.6.3 The unitSize attribute
The unitSize attribute is of type double and represents the relative size of the Compartment with respect to the size of the Domains to which they are mapped. Thus for any infinitesimal subset of the Domain with size S, there exists an amount of Compartmenti of size (S*unitSizei) for i=1..N compartments mapped to that DomainType. For example, a 3D Compartment (and DomainType) which is mapped to a 3D DomainType has a unitSize which is a volume fraction of dimensionless unit. The total set of all such volume fractions mapped to a particular DomainType will typically sum to one.
If the spatialDimensions attribute of the parent Compartment is different than the spatialDimensions attribute of referenced DomainType, the unitSize attribute is a conversion factor between the two. The most common example of this would be a 2D Compartment being mapped to a 3D DomainType, such as an ER-membrane being mapped to a volumetric cell interior. In this case, the unitSize is a surface-to-volume ratio. (edited) 
```

so it has to be between 0 and 1 if the dimensions match, otherwise it would be a conversion factor that may well be bigger than 1.

fbergmann:
## Description
before testing whether the unitsize is between 0 and 1, we should check, that the spatial dimensions match first. 

## Motivation and Context
fixes #275

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/packages/spatial/validator/SpatialSBMLErrorTable.h
+++ b/src/sbml/packages/spatial/validator/SpatialSBMLErrorTable.h
@@ -884,7 +884,7 @@ static const packageErrorTableEntry spatialErrorTable[] =
     LIBSBML_CAT_GENERAL_CONSISTENCY,
     LIBSBML_SEV_WARNING,
     "The attribute 'spatial:unitSize' on a <compartmentMapping> must have a value "
-    "between 0 and 1, inclusive.",
+    "between 0 and 1, inclusive, when the dimensions of the referenced compartments have the same dimensions.",
     { "L3V1 Spatial V1 Section"
     }
   },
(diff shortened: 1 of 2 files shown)


luciansmith:
We can change the wording to whatever, but it should match https://github.com/sbmlteam/sbml-specifications/issues/392

luciansmith:
@fbergmann : We also need to apply the caveat to rule 21351.  I looked at the code, but I got stymied; can you look at it?  It's the function:

SpatialCompartmentMappingUnitSizesCheck::check_

skeating (approved):

luciansmith:
We still need checks for rule 21351, not just 21350.

fbergmann:
> @fbergmann : We also need to apply the caveat to rule 21351. I looked at the code, but I got stymied; can you look at it? It's the function:
> 
> SpatialCompartmentMappingUnitSizesCheck::check_

@luciansmith that code should probably be adjusted such that it only sums up the compartmentmappings, for which their compartmetns spatialDimensions match. Though i'm unsure of how that fix would actually look like. Since it is a warning, it didnt stop the exchange between tools. 

Maybe one idea could be to change the check such that it would only try and sum up fractional unitSizes < 1. As that check was probably be intended to tell you when you added too many fractions.

PR #284, opened 2022-11-29
Fix crash when infering units

avandecreme:
## Description

When trying to infer units of the added `inferUnits-4.xml` file, libsbml crashes.
My change only fixes the crash but does not make libsbml able to infer the units.

The reason why libsbml crashes (or is not able to infer the units) is that the file contains a `function_1` with more than 2 arguments.  
`UnitFormulaFormatter::inferUnitDefinition` then tries to reduce the kinetic law's math node to a binary tree but fails to do that properly because of the `function_1` call.
Maybe a full fix would be to fix `ASTNode::reduceToBinary()` to detect that there is a call to a function of more than 2 params and build a binary tree of the function first.
For now, avoiding a crash seems like a quick win.

Note that replacing `<ci> function_1 </ci>` by `<time />` in the input file avoids the crash and let libsbml infer units correctly.

## Motivation and Context

Avoid crashing when infering units.

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/units/UnitFormulaFormatter.cpp
+++ b/src/sbml/units/UnitFormulaFormatter.cpp
@@ -377,7 +377,10 @@ UnitFormulaFormatter::getUnitDefinitionFromFunction(const ASTNode * node,
 
   if(node->getType() == AST_FUNCTION)
   {
-    const FunctionDefinition *fd = 
+    // The function name get lost by ASTNode::reduceToBinary when it has more
+    // than one parameter. So we need to check we have a name before calling
+    // model->getFunctionDefinition to avoid crashing.
+    const FunctionDefinition *fd = node->getName() == NULL ? NULL :
                                model->getFunctionDefinition(node->getName());
     if (fd && fd->isSetMath())
     {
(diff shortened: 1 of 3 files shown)


src/sbml/units/UnitFormulaFormatter.cpp
```diff
@@ -377,7 +377,7 @@ UnitFormulaFormatter::getUnitDefinitionFromFunction(const ASTNode * node,
 
   if(node->getType() == AST_FUNCTION)
   {
-    const FunctionDefinition *fd = 
+    const FunctionDefinition *fd = node->getName() == NULL ? NULL :
```
eltix, line 383: maybe add a comment here explaining why `getName()` is NULL when the function takes more than 2 arguments? Because it is not clear judging by the function name only.

eltix:
cc @fbergmann

fbergmann (approved):
good catch! thank you for the submission.

avandecreme:
Thanks for reviewing and merging @fbergmann!
Should I open an issue regarding making unit inference work on this case?

fbergmann:
yes please. For inference in your local work, i recommend that you'd first run the inline function definition converter on the model prior to inferring units. That workaround will make it work for you.

PR #300, opened 2023-03-03
Add test case for replaceFD issue #299

lkeegan:
## Description
test case for replaceFD issue #299

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Change in documentation

## Checklist:

- [ ] I have updated all documentation necessary.
- [ ] I have checked spelling in (new) comments.

## Testing
- [X] Testing is done automatically and codecov shows test coverage
- [ ] This cannot be tested automatically

Issue #299, opened 2023-03-03
bug in `SBMLTransforms::replaceFD`?


lkeegan:
I was using `SBMLTransforms::replaceFD` to expand/inline functions in math expressions but ran into incorrect results in some cases:

https://github.com/spatial-model-editor/spatial-model-editor/issues/856#issuecomment-1451919941

I think substituting each argument sequentially like this is only valid if there is no overlap between the variables in the function body and the variables that are passed to the function?

fbergmann:
This has now been fixed and merged into development. We plan to have a new release including the fix within the next couple of weeks.

lkeegan:
@fbergmann thanks! Unfortunately while this fixes the failing test case I don't think it is correct in general, e.g. changing the expression from `p * S1` to `p - S1` gives incorrect results. I think to do the substitution correctly it needs to be done for all variables at the same time - I've made a PR with an attempt at doing this here: #303

PR #303, opened 2023-03-14
Fix bug in `SBMLTransforms::replaceFD`

lkeegan:
## Description

- add `ASTNode::replaceArguments` which replaces all supplied arguments at the same time
- modify `SBMLTransforms::replaceBvars` to use this instead of replacing arguments sequentially
- refactor shared functionality into `copyNode`
- modify test case from using mul to minus to show some results are still incorrect after #300

## Motivation and Context

`SBMLTransforms::replaceFD` gave incorrect results in some cases.
This was due to `SBMLTransforms::replaceBvars` replacing each bvar sequentially (#299)
This PR instead substitutes all bvars simultaneously to avoid these incorrect results.

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


fbergmann (approved):
that does look better, thanks

PR #310, opened 2023-03-25
Xmlns in math

skeating:
## Description

adds multi ns to math if it is declared
will also work for any packages
## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/examples/c++/echoSBML.cpp
+++ b/examples/c++/echoSBML.cpp
@@ -47,15 +47,13 @@ LIBSBML_CPP_NAMESPACE_USE
 int
 main (int argc, char* argv[])
 {
-  //if (argc != 3)
-  //{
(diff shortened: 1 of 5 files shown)


fbergmann (changes requested):
I suspect we are leaking the new namespaces: 

* on setDeclared... we clone the incoming namespace
* in assignment operator and copy ctor we steal the namespace from the rhs
* in destructor we dont free it. 

the also probably should be an unsetDeclared... function to get rid / free it. as well

fbergmann (changes requested):
`unsetDe...` needs to set mNamespaces to NULL after deleting, as otherwise it could be called again, resulting in a double free.

fbergmann (approved):

PR #317, opened 2023-04-07
Fix namespace conflict on math nodes

### LINKED ISSUE #316: Conversion failure in v5.20.0
When running `./examples/c/convertSBML` on [this unzipped file](https://github.com/sbmlteam/libsbml/files/11171770/Dextromethorphan.zip) I get the following error in v5.20.0:

```
Attempting to convert model to SBML Level 3 Version 2.
Unable to perform conversion due to the following:
line 4950: (01010 [Error]) Duplicate XML attribute.

Conversion skipped.  Either libSBML does not (yet) have
ability to convert this model, or (automatic) conversion
is not possible in this case.
```
It is working fine in v5.19

Here is the line 4950 mentioned:
`        <math xmlns="http://www.w3.org/1998/Math/MathML" xmlns:sbml="http://www.sbml.org/sbml/level3/version1/core">`

I suspect the issue is introduced by #310 and more specifically the following lines which copy existing namespaces without checking for collision with the one added a few lines above:
https://github.com/sbmlteam/libsbml/pull/310/files#diff-a1350eea6646b5dc5f2a33f27c7fed1ccb413304f0d4a6df3909a2f0ec15f0c2R2232-R2245

avandecreme:
## Description
This is an attempt to fix #316. I am not super happy with that fix because it effectively makes the `sbml` namespace reserved but I don't have a better idea.

## Motivation and Context

Fix #316

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/conversion/test/TestLevelVersionConverter.cpp
+++ b/src/sbml/conversion/test/TestLevelVersionConverter.cpp
@@ -461,6 +461,32 @@ START_TEST (test_specref_ref_real)
 END_TEST
 
 
+START_TEST (test_convertToL3V2)
+{
+  SBMLNamespaces sbmlns(3, 2);
+  ConversionProperties prop(&sbmlns);
+  prop.addOption("strict", true, "should validity be preserved");
+  prop.addOption("setLevelAndVersion", true, "convert the document to the given level and version");
+  prop.addOption("ignorePackages", true);
+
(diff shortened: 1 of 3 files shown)


fbergmann (approved):
thank you for this. We should indeed not add two sbml namespaces.

Issue #319, opened 2023-05-03
libsbml writing out empty <listOfReactants/>


luciansmith:
This is a change of behavior from not long ago:  if I flatten the attached model, I now get <listOfReactants/> and didn't use to.  Is this behavior intentional?

[deleteReactant.xml.txt](https://github.com/sbmlteam/libsbml/files/11391780/deleteReactant.xml.txt)

luciansmith:
Correction!  This is a difference in behavior of an L3v1 model vs. an L3v2 model, because of Reaction::writeElements, which checks

```
    if (mReactants.hasOptionalElements() == true ||
        mReactants.hasOptionalAttributes() == true ||
        mReactants.isExplicitlyListed())
    {
      mReactants.write(stream);
    }
```

Is there an obvious place to turn off isExplicitlyLIsted in the flattening routines, assuming we need to do that there?

fbergmann:
There is 

```c++

ListOf::setExplicitlyListed(bool value)

```

so you should be able to set it on all lists you want during the flattening procedure.

skeating:
Lucian sorted this

PR #321, opened 2023-05-05
Fix comp flattening leaving behind empty lists.

luciansmith:
* Add ListOf elements to getAllElements when explicitly listed is 'true'.
* Use this to clear the explicitlyListed flag when flattening comp models, so deleted elements won't leave empty lists behind for l3v2.


luciansmith:
Fix for #319

PR #326, opened 2023-05-09
Various small fixes

### LINKED ISSUE #37: bug in defaultvalues
The build log also noted an unsafe return here: 

    2020-08-10T08:30:45.2176030Z   /tmp/pip-req-build-q_svt3q_/libsbml_source/src/sbml/packages/render/sbml/DefaultValues.cpp:3775:18: warning: returning address of local temporary object [-Wreturn-stack-address]
    2020-08-10T08:30:45.2178260Z     return (char *)(dv->getFillRuleAsString()).c_str();
    2020-08-10T08:30:45.2178980Z                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~
    2020-08-10T08:30:47.3440010Z   1 warning generated.

### LINKED ISSUE #134: c api catches exception by value rather than by reference
Compiling libSBML with manylinux2010, i see a lot of warnings of the form: 

```
./base/sbml/ModifierSpeciesReference.cpp:435:10: warning: catching polymorphic type ‘class SBMLConstructorException’ by value [-Wcatch-value=]
   catch (SBMLConstructorException)
          ^~~~~~~~~~~~~~~~~~~~~~~~
./base/sbml/ModifierSpeciesReference.cpp: In function ‘ModifierSpeciesReference_t* ModifierSpeciesReference_createWithNS(SBMLNamespaces_t*)’:
./base/sbml/ModifierSpeciesReference.cpp:451:10: warning: catching polymorphic type ‘class SBMLConstructorException’ by value [-Wcatch-value=]
   catch (SBMLConstructorException)
          ^~~~~~~~~~~~~~~~~~~~~~~~
```

we should change this to catch by reference.

### LINKED ISSUE #149: SBMLWriter crashes on NULL document
the SBML writer fails to check the given document pointer, and will dereference it if NULL.

### LINKED ISSUE #324: Typo in rateOf error message
There should be a whitespace in the following error message between species and 'atp'.
```
The formula 'rateOf(atp)' in the math element of the <assignmentRule> uses the species'atp' whose compartment is referenced as the variable in an assignmentRule.
```

### LINKED ISSUE #325: Validation of SBML-comp models
Validation of an SBML-comp model (SBML semantic test suite case 01161) does not give any validation errors/warnings initially, but after applying `CompFlatteningConverter` it does. The warnings seem unrelated to flattening the model. Why aren't they shown before? Are those checks skipped if `comp` is used? 

To reproduce via python-libsbml:
```python
import libsbml
import requests

url = "https://raw.githubusercontent.com/sbmlteam/sbml-test-suite/release/cases/semantic/01161/01161-sbml-l3v2.xml"
r = requests.get(url, allow_redirects=True)

sbml_reader = libsbml.SBMLReader()
sbml_doc = sbml_reader.readSBMLFromString(r.content.decode("utf-8"))
assert sbml_doc.validateSBML() == 0
for i_error in range(sbml_doc.getNumErrors()):
    error = sbml_doc.getError(i_error)
    print(f'{error.getCategoryAsString()} ({error.getSeverityAsString()}): {error.getMessage()}')

conversion_properties = libsbml.ConversionProperties()
conversion_properties.addOption("flatten comp", True)
conversion_properties.addOption("leave_ports", False)
conversion_properties.addOption("abortIfUnflattenable", "none")
assert sbml_doc.convert(conversion_properties) == libsbml.LIBSBML_OPERATION_SUCCESS
str_model = libsbml.SBMLWriter().writeSBMLToString(sbml_doc)

print(str_model)

n_errors = sbml_doc.validateSBML()
for i_error in range(sbml_doc.getNumErrors()):
    error = sbml_doc.getError(i_error)
    print(f'{error.getCategoryAsString()} ({error.getSeverityAsString()}): {error.getMessage()}')

# FAILS
assert n_errors == 0, n_errors
```

### LINKED ISSUE #328: Document NULL for SBMLWriter test crashes on mac

skeating:
## Description

Various small fixes from issues

## Motivation and Context

Fixes #149
Fixes #324 
Fixes #37 
Fixes #325 
Fixes #328
Fixes #134

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/AlgebraicRule.cpp
+++ b/src/sbml/AlgebraicRule.cpp
@@ -371,7 +371,7 @@ AlgebraicRule_create(unsigned int level, unsigned int version)
     AlgebraicRule* obj = new AlgebraicRule(level,version);
     return obj;
   }
-  catch (SBMLConstructorException)
+  catch (SBMLConstructorException &e)
   {
     return NULL;
   }
@@ -387,7 +387,7 @@ AlgebraicRule_createWithNS (SBMLNamespaces_t* sbmlns)
     AlgebraicRule* obj = new AlgebraicRule(sbmlns);
     return obj;
   }
-  catch (SBMLConstructorException)
+  catch (SBMLConstructorException &e)
   {
     return NULL;
   }
--- a/src/sbml/AlgebraicRule.cpp
+++ b/src/sbml/AlgebraicRule.cpp
@@ -371,7 +371,7 @@ AlgebraicRule_create(unsigned int level, unsigned int version)
     AlgebraicRule* obj = new AlgebraicRule(level,version);
     return obj;
   }
-  catch (SBMLConstructorException &e)
+  catch (SBMLConstructorException &)
   {
     return NULL;
   }
@@ -387,7 +387,7 @@ AlgebraicRule_createWithNS (SBMLNamespaces_t* sbmlns)
     AlgebraicRule* obj = new AlgebraicRule(sbmlns);
     return obj;
   }
-  catch (SBMLConstructorException &e)
+  catch (SBMLConstructorException &)
   {
     return NULL;
   }
--- a/src/sbml/packages/render/sbml/DefaultValues.cpp
+++ b/src/sbml/packages/render/sbml/DefaultValues.cpp
@@ -3777,7 +3777,7 @@ LIBSBML_EXTERN
 char *
 DefaultValues_getFillRuleAsString(const DefaultValues_t * dv)
 {
-  return (char *)(dv->getFillRuleAsString()).c_str();
+  return (char *)(FillRule_toString((FillRule_t)(dv->getFillRule())));
 }
 
 
--- a/src/sbml/packages/render/sbml/DefaultValues.cpp
+++ b/src/sbml/packages/render/sbml/DefaultValues.cpp
@@ -3777,7 +3777,7 @@ LIBSBML_EXTERN
 char *
 DefaultValues_getFillRuleAsString(const DefaultValues_t * dv)
 {
-  return (char *)(FillRule_toString((FillRule_t)(dv->getFillRule())));
+  return (char *)safe_strdup(FillRule_toString((FillRule_t)(dv->getFillRule())));
 }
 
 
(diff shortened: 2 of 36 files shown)


src/sbml/packages/render/sbml/DefaultValues.cpp
```diff
@@ -3777,7 +3777,7 @@ LIBSBML_EXTERN
 char *
 DefaultValues_getFillRuleAsString(const DefaultValues_t * dv)
 {
-  return (char *)(dv->getFillRuleAsString()).c_str();
+  return (char *)(FillRule_toString((FillRule_t)(dv->getFillRule())));
```
fbergmann: @skeating this is still invalid. Either it would have to be returned as `const char*`, or we need a `strdup`, as the resulting ptr is not owned by the caller otherwise.

src/sbml/AlgebraicRule.cpp
```diff
@@ -371,7 +371,7 @@ AlgebraicRule_create(unsigned int level, unsigned int version)
     AlgebraicRule* obj = new AlgebraicRule(level,version);
     return obj;
   }
-  catch (SBMLConstructorException)
+  catch (SBMLConstructorException &e)
```
fbergmann: @skeating instead of defining a variable that we don't use, would it not be better to just change the type to `SBMLConstructorException &` without defining the variable?

fbergmann:
I still think the best change would be to not have the constructors throw exceptions at all (that does not help anyone IMO). But if we cannot change that, I think it would be better not to introduce the variables (`e`) that we don't use. What do you think?

src/sbml/AlgebraicRule.cpp
luciansmith: Variables you don't use do produce warnings, and the fewer of those the better. But wouldn't passing on the error message somehow be better still?

fbergmann: in this case, here it is a c-function, that is trying to create something with invalid arguments, so returning `NULL` would be the expected result. I certainly wouldn't want to have messages printed anywhere. 

In the most cases where this is caught it is from internal ::createXXX functions, that take level / version from the SBMLDocument. So something would already have been very wrong long before the create call for this exception to ever be raised.

fbergmann (approved):
thank you

PR #331, opened 2023-05-26
Including CMake modules for some find-dependency-.cmake scripts

adelhpour:
## Description
Some CMake models are induced in find-dependency-.cmake scripts

## Motivation and Context
While linking against the libsbml, there was an error with finding the dependencies.

## Types of changes
include(CheckLibraryExists) is added to FindBZ2.cmake
include(CheckCSourceCompiles) is added to FindLIBXML.cmake


fbergmann (approved):
thank you for the PR adel, this looks good to me

PR #335, opened 2023-06-08
Conversion/validation fixes

### LINKED ISSUE #327: checkConsistency does not report undefined function from within another function
Calling `checkConsistency` on a document will report references to undefined functions, for example with the following file:
```xml
<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" level="3" version="2">
  <model timeUnits="second">
    <listOfCompartments>
      <compartment id="compartment" units="dimensionless" constant="false"/>
    </listOfCompartments>
    <listOfRules>
      <assignmentRule variable="compartment">
        <math xmlns="http://www.w3.org/1998/Math/MathML">
          <apply>
            <root/>
            <apply>
              <ci> undefined_function </ci>
            </apply>
          </apply>
        </math>
      </assignmentRule>
    </listOfRules>
  </model>
</sbml>
```

Running `validateSBML` will output:
```
line 8: (10214 [Error]) Outside of a <functionDefinition>, if a <ci> element is the first element within a MathML <apply>, then the <ci>'s value can only be chosen from the set of identifiers of <functionDefinition>s defined in the SBML model.
Reference: L3V2 Section 4.3.2
 The formula 'undefined_function()' in the math element of the <assignmentRule> uses 'undefined_function' which is not a function definition id.
```

However, if a `functionDefinition` references an undefined function, no such error is reported. Here is an example file:
```xml
<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" level="3" version="2">
  <model timeUnits="second">
    <listOfFunctionDefinitions>
      <functionDefinition metaid="_tt" id="fd" name="ggh">
        <math xmlns="http://www.w3.org/1998/Math/MathML">
          <lambda>
            <bvar>
              <ci> x </ci>
            </bvar>
            <apply>
              <power/>
              <ci> x </ci>
              <apply>
                <ci> undefined_function </ci>
                <cn> 0.5 </cn>
              </apply>
            </apply>
          </lambda>
        </math>
      </functionDefinition>
    </listOfFunctionDefinitions>
  </model>
</sbml>
```

What do you think about checking function definitions as well for references to undefined functions?

### LINKED ISSUE #333: Segmentation fault with some operations after failed strict conversion
The following program seg fault when run with the [following uncompressed file](https://github.com/sbmlteam/libsbml/files/11601071/conversion_seg_fault.zip).

```c
printParentDocIsNull(SBMLDocument_t * d)
{
  Model_t *m = SBMLDocument_getModel(d);
  SBMLDocument_t *d2 = SBase_getSBMLDocument((SBase_t *)m);
  printf("Parent doc is null: %i\n", d2 == NULL);
}


int
main (int argc, char *argv[])
{
  int success;
  SBMLDocument_t *d = readSBML(argv[1]);
  printParentDocIsNull(d);
  success = SBMLDocument_setLevelAndVersionStrict(d, 3, 2);
  printf("Strict Success: %d\n", success);
  printParentDocIsNull(d);
  success = SBMLDocument_setLevelAndVersionNonStrict(d, 3, 2);
  printf("Non Strict Success: %d\n", success);
  printParentDocIsNull(d);
  SBMLDocument_checkConsistency(d);

  SBMLDocument_free(d);
  return 0;
}
```

Result in:
```
Parent doc is null: 0
Strict Success: 0
Parent doc is null: 1
Non Strict Success: 1
Parent doc is null: 1
Segmentation fault (core dumped)
```

I tracked down the issue to the fact when failing to do a strict conversion, we restore the model of the document to a stored copy here:
https://github.com/sbmlteam/libsbml/blob/49ce5b40ead6862c1a610a569d0606f03a4fb4fa/src/sbml/conversion/SBMLLevelVersionConverter.cpp#L424

However, copying a model does not copy the `mSBML` field on `SBase`:
https://github.com/sbmlteam/libsbml/blob/49ce5b40ead6862c1a610a569d0606f03a4fb4fa/src/sbml/SBase.cpp#L421
This is put in evidence by the `Parent doc is null` outputs above.

This then cause the crash when running the consistency check on unique ids because the following line trust that `m.getSBMLDocument()` will return a non null value:
https://github.com/sbmlteam/libsbml/blob/49ce5b40ead6862c1a610a569d0606f03a4fb4fa/src/sbml/validator/constraints/UniqueIdsInModel.cpp#L166

I am able to fix that specific issue by adding the following line when cloning the model in the `SBMLLevelVersionConverter`:
```cpp
          currentModel->connectToParent(mDocument);
```
However, I am not sure this is the best approach because all other `SBase` (reactions, compartments...) of the model will still be missing there back link to the document.

Any idea how to best fix this?

### LINKED ISSUE #334: Segmentation fault when doing function expansion and unit inference after conversion to L3V2
The following program segfault when fed the [following uncompressed file](https://github.com/sbmlteam/libsbml/files/11624961/many-params-function.zip)

```c
#include <stdio.h>
#include <sbml/SBMLTypes.h>
#include "util.h"

int main (int argc, char *argv[])
{
  if (argc != 2)
  {
    printf("Usage: readSBML filename\n");
    return 2;
  }

  const char * filename = argv[1];
  SBMLDocument_t *d = readSBML(filename);

  int success = SBMLDocument_setLevelAndVersion(d, 3, 2);
  printf("Success: %d\n", success);

  ConversionProperties_t* props;

  props = ConversionProperties_create();
  ConversionOption_t * expandFuncDefOpt = ConversionOption_create("expandFunctionDefinitions");
  ConversionProperties_addOption(props, expandFuncDefOpt);
  SBMLDocument_convert(d, props);
  ConversionOption_free(expandFuncDefOpt);
  ConversionProperties_free(props);

  props = ConversionProperties_create();
  ConversionOption_t * inferUnitsOpt = ConversionOption_create("inferUnits");
  ConversionProperties_addOption(props, inferUnitsOpt);
  SBMLDocument_convert(d, props);
  ConversionOption_free(inferUnitsOpt);
  ConversionProperties_free(props);

  SBMLDocument_free(d);
}
```

From what I gathered, it is because `tempUD1` is set to `NULL` here:
https://github.com/sbmlteam/libsbml/blob/49ce5b40ead6862c1a610a569d0606f03a4fb4fa/src/sbml/units/UnitFormulaFormatter.cpp#L2635
The reason for that is that `inverseFunctionOnUnits` calls ` UnitDefinition::divide` on two `UnitDefinition`s having different level and version (L2V4 like the original doc and L3V2 like the converted one), so we are falling in the following branch:
https://github.com/sbmlteam/libsbml/blob/49ce5b40ead6862c1a610a569d0606f03a4fb4fa/src/sbml/UnitDefinition.cpp#L1467

I found two ways to avoid the crash:
1. move the level/version conversion after function expansion and unit inference
2. keep level/version conversion first but group function expansion and unit inference into one convert call like this:
```c
  props = ConversionProperties_create();
  ConversionOption_t * expandFuncDefOpt = ConversionOption_create("expandFunctionDefinitions");
  ConversionProperties_addOption(props, expandFuncDefOpt);
  ConversionOption_t * inferUnitsOpt = ConversionOption_create("inferUnits");
  ConversionProperties_addOption(props, inferUnitsOpt);
  SBMLDocument_convert(d, props);
  ConversionOption_free(expandFuncDefOpt);
  ConversionOption_free(inferUnitsOpt);
  ConversionProperties_free(props);
```

Having level/version conversion in between function expansion and unit inference also segfault.

skeating:
## Motivation and Context

Fixes #327 
Fixes #333 
Fixes #334 

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


src/sbml/conversion/SBMLLevelVersionConverter.cpp
```diff
@@ -1321,6 +1313,32 @@ SBMLLevelVersionConverter::collectSpeciesReferenceIds()
 
   return srids;
 }
+
+
+void
+SBMLLevelVersionConverter::updateUnitMap(SBMLDocument* d)
+{
+  if (d->getModel()->isPopulatedListFormulaUnitsData())
```
fbergmann: it would be good to check whether a model is set (should be the case, but since this is a function we might want to check)

fbergmann (approved):

PR #342, opened 2023-08-19
add a C api for SBMLErrorLog

### LINKED ISSUE #329: Give access to document error log in C API
It would be nice to have access to the error log from the C API.

My use case is that I might get some errors from some checks/conversion. I will then fix them and re-run those checks/conversion.
If I do that, calling `SBMLDocument_getNumErrors(doc)` will count the errors from both checks/conversion runs whereas I only want the ones from the second run.

For most checks, I can rely on the number of errors returned by the check and compute the offset of the first relevant error.
However, the conversion returns a boolean instead of a number of errors so I cannot do that. I still found a workaround in that case, the level and version converter actually clears the log (except in case of early return) so I can set the offset to 0.

Still, I think that being able to clear the log from the C API would be nice. While there is an API to clear the log, there is no way to retrieve the log object from a document.

So my request would be to add something like:
```
LIBSBML_EXTERN
const SBMLErrorLog_t *
SBMLDocument_getErrorLog(SBMLDocument_t *d)
{
  return (d != NULL) ? d->getErrorLog() : NULL;
}
```

I am not sure about `SBMLErrorLog_t` since it doesn't exist at the moment. Returning `XMLErrorLog_t` instead would be fine for my use case.

What do you think?

skeating:
## Description

adds a c api for SBMLErrorLog

## Motivation and Context

Fixes #329 
## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/SBMLErrorLog.cpp
+++ b/src/sbml/SBMLErrorLog.cpp
@@ -397,7 +397,7 @@ SBMLErrorLog::getError (unsigned int n) const
  * @return the nth SBMLError in this log.
  */
 const SBMLError*
-SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
+SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity, unsigned int fromCAPI) const
 {
   unsigned int count = 0;
   MatchSeverity matcher(severity);
@@ -406,7 +406,17 @@ SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
   {
     if (matcher(*it))
     {
-      if (count == n) return dynamic_cast<const SBMLError*>(*it);
+      if (count == n)
+      {
+        if (fromCAPI == 0)
+        {
+          return dynamic_cast<const SBMLError*>(*it);
+        } 
+        else
+        {
+          return static_cast<const SBMLError*>(*it);
+        }
+      }
       ++count;
     }
   }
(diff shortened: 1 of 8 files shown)


src/sbml/SBMLErrorLog.cpp
```diff
@@ -406,7 +406,17 @@ SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
...
+      {
+        if (fromCAPI == 0)
+        {
+          return dynamic_cast<const SBMLError*>(*it);
+        } 
+        else
+        {
+          return static_cast<const SBMLError*>(*it);
+        }
+      }
```
fbergmann: this looks scary, is this really needed? Is it safe? because those could always be XMLErrors and not SBMLErrors, so when accessing them from the C API we'd be giving them bad pointers. It would be better to just use the get error function returning an XMLError and have a function to figure out whether an XMLError is an SBMLError (which would be clear from the error code)

I'm sure I could devise a way to make that static cast do bad things if needed. So let's think about this once again. (sorry)

src/sbml/SBMLErrorLog.cpp
```diff
@@ -417,6 +427,104 @@ SBMLErrorLog::getErrorWithSeverity(unsigned int n, unsigned int severity) const
...
+
+
+LIBSBML_EXTERN
+const SBMLError_t *
+SBMLErrorLog_getError(const SBMLErrorLog_t *log, unsigned int n)
+{
+  if (log == NULL) 
+    return NULL;
+
+  return (const SBMLError_t*)(XMLErrorLog_getError((const XMLErrorLog_t*)(log), n));
```
fbergmann: I think we should still return a dynamic cast to an SBMLError, since the error returned might not be an SBMLError. At least we should have a test for this.

src/sbml/test/TestSBMLErrorLog.c
```diff
@@ -0,0 +1,168 @@
...
+  fail_unless(SBMLErrorLog_getNumFailsWithSeverity(log, 1) == 2);
+
+  XMLError_free(e);
+  XMLError_free(e1);
+  //XMLError_free((SBMLError_t*)(r));
+  //XMLError_free((SBMLError_t*)(r1));
+  SBMLErrorLog_free(log);
+}
+END_TEST
+
```
fbergmann, line 161: I think we need tests that test retrieving SBMLErrors when they are XMLErrors ensuring we don't garble ptrs.

fbergmann (changes requested):
hello Sarah, I added some remarks, could you just verify that I'm totally wrong on those. thank you.

src/sbml/SBMLErrorLog.cpp
skeating: ok reverted my static casts

skeating:
okay  I changed the static cast  and added test for the different types of error. I must admit I'm a little bit confused by the results

fbergmann (approved):
thank you for these changes, this looks great

PR #346, opened 2023-09-06
Make various 'get' functions const for annotations.

luciansmith:
I was trying to 'getYear' on a const Date, and couldn't.

'ModelCreator' get functions were already const.


fbergmann (approved):
looks good to me, thanks

PR #362, opened 2024-01-18
Allow FBC objects to be used in both L3v1 and L3v2 core documents.

### LINKED ISSUE #360: FBC:  Can't make package version 3 object in L3v2 core model.
It's impossible to create a flux objective and add it to an L3v2 model:

* If you run createFluxObjective(), it gives you an object with a getLevel() of 3, a getVersion() of 2, and a getPackageVersion() of 3.  When you call 'setVariableType', it notices that getVersion is 2, and returns LIBSBML_UNEXPECTED_ATTRIBUTE.
* If you create a FluxObjective with FluxObjective(3,1,3), it lets you call 'setVariableType', but doesn't let you call addFluxObjective(), because of the version mismatch to the core SBML object.

I propose that packages in general should never check the core level or version; they should only check the package version.  (Perhaps the layout/render functions could check the core level, but even they shouldn't check the core version.)  Putting 'version1' in the namespace string turns out to be a terrible idea, but that ship has sailed; at this point, all we can do is just not check it.

For the immediate problem, I say we don't check versions when calling 'addXXX', nor when calling 'setVariableType' for FBC objects, but a more thorough scouring of the code will probably be necessary to check for other packages moving forward.

@skeating @fbergmann Opinions?

luciansmith:
Fixes #360 

Some objects already didn't check the version, and some didn't check the level, either.  This standardizes on just checking the package version.

Doesn't fix anything for other packages, just FBC.

### DIFF
--- a/src/sbml/packages/fbc/extension/FbcModelPlugin.cpp
+++ b/src/sbml/packages/fbc/extension/FbcModelPlugin.cpp
@@ -656,10 +656,6 @@ int
   {
     return LIBSBML_LEVEL_MISMATCH;
   }
-  else if (getVersion() != bound->getVersion())
-  {
-    return LIBSBML_VERSION_MISMATCH;
-  }
   else if (getPackageVersion() != bound->getPackageVersion())
   {
     return LIBSBML_PKG_VERSION_MISMATCH;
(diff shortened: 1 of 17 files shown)


fbergmann (approved):
To be honest, I would prefer libSBML would do way less checking before setting the things that API calls want to set. And I don't think rejecting API calls should ever be made. 

The changes you propose work fine.

thanks

PR #366, opened 2024-02-19
Enable CMake user option `BUILD_SHARED_LIB`

Waschina:
## Description
The CMake configuration does not respect the user option "-DBUILD_SHARED_LIB=<ON/OFF>". Package managers (and users) commonly use this main configuration option to decide whether to build the project as a shared or static library. Specifically, library build systems for dependencies of R-packages expect the `BUILD_SHARED_LIB` option: ([Windows build system](https://github.com/r-windows/rtools-packages/blob/31f09c0801f9b038e91e8905ad83473a7c447472/mingw-w64-libsbml/PKGBUILD), [MacOS build system](https://github.com/R-macos/recipes/pull/49)).

Generally, the `BUILD_SHARED_LIB` variable directly affects the command `add_library()` if no type (`SHARED` or `STATIC`) is given to the function. Instead of `BUILD_SHARED_LIB`, libsbml's CMake setup seems to use the non-canonical options `LIBSBML_SKIP_SHARED_LIBRARY` and `LIBSBML_SKIP_STATIC_LIBRARY`:

https://github.com/sbmlteam/libsbml/blob/a3e217851f18d3c5e8e2b0082fb98176738491e8/src/CMakeLists.txt#L512C1-L513C2
https://github.com/sbmlteam/libsbml/blob/a3e217851f18d3c5e8e2b0082fb98176738491e8/src/CMakeLists.txt#L462C1-L464C2

## Motivation and Context
Although I think the libsbml-specific shared/static build config options (LIBSBML_SKIP_<SHARED/STATIC>_LIBRARY) are totally fine, I would suggest enabling the widespread user option `BUILD_SHARED_LIB` as well, which would make libsbml also more accessible and easier to implement for developers of packages and toolchains.

## Types of changes
- [ ] Change in documentation

## Checklist:

- [X] I have checked spelling in (new) comments.

## Testing
- [X] This cannot be tested automatically
I tested the new option manually using the CMake GUI in Windows. It worked as expected: Shared library build can be skipped with the new option "-DBUILD_SHARED_LIB=OFF".

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -171,10 +171,20 @@ set (LIBSBML_FIND_MODULES "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules/FindLIBSBML.
 #
 
 # Build static / shared library
-option(BUILD_SHARED_LIBS  "Build shared library. (Set to OFF to build static libraries.)" ON)
+# 
+# Here we check whether the user has defined BUILD_SHARED_LIBS. If it is
+# defined we will honor the value, otherwise we continue to use the 
+# LIBSBML_SKIP variables to determine whether to build shared or static
+# libraries.
 
-if(NOT BUILD_SHARED_LIBS)
-  set(LIBSBML_SKIP_SHARED_LIBRARY ON)
+if(DEFINED BUILD_SHARED_LIBS)
+  if (BUILD_SHARED_LIBS)
+    set(LIBSBML_SKIP_SHARED_LIBRARY OFF)
+    set(LIBSBML_SKIP_STATIC_LIBRARY ON)
+  else()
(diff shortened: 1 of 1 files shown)


fbergmann (commented):
Thanks for the PR, I agree we should honor the BUILD_SHARED_LIBS option if specified. But we should not set a default. I'll make slight changes so that 

* if BUILD_SHARED_LIBS is defined and TRUE, only the shared lib is built, 
* if it is defined and FALSE only the static one is built 
* if not defined the build behavior is the same as before (building both by default, or honoring LIBSBML_SKIP variables)

that should solve your issue, and not change the behavior for anyone else.

fbergmann (approved):

PR #370, opened 2024-04-16
Replace deprecated unittest.makeSuite()

penguinpee:
The function has been deprecated in Python 3.11 and will be removed from
the upcoming Python 3.13.

See: https://docs.python.org/3.13/whatsnew/3.13.html#unittest


penguinpee:
Fixed the `TypeError` in `test_python_binding` by first instantiating the class.

fbergmann (approved):
thank you so much for this PR!

PR #388, opened 2024-08-19
Remove the check for zlib.h.

luciansmith:
When co-compiling zlib and libsbml, zlib.h won't exist when checked by CMake, but it'll exist soon enough.


fbergmann (approved):
i've moved the library check also to the find module. And removed the comment testing for the presence of the zlib header. (no use leaving it just commented out)

PR #393, opened 2024-08-27
Issue 385: flexibility with running strict unit validation

### LINKED ISSUE #385: checkConsistencyWithStrictUnits reports Units issues as Errors not Warnings
Hi all,
this is related to:
https://github.com/sbmlteam/libsbml/issues/378

I switched from `checkConsistency` to `checkConsistencyWithStrictUnits` and now get a lot of errors instead of warnings.
My understanding is that the function is just performing some additional more strict rules for unit validation to handle numerical conversion factors based on the documentation:
 
> "Performs consistency checking and validation on this SBML document using the ultra strict units validator that assumes that there are no hidden numerical conversion factors."

But in addition the function reports units consistency issues as "error" not as "warning". This is incorrect. Despite the more strict checking the unit consistency issues are still warnings and the model is valid! 
The unit consistency issues have to be reported as warnings, so that `checkConsistency` to `checkConsistencyWithStrictUnits` behave in the same way (with just additional tests in the strict version) and the validation result is in line with "units being annotations but not making the model invalid".

Best Matthias

fbergmann:
## Description
This PR adds the strict unit validator to the options, so it can be run using checkConsistency. Additionally it adds an override flag to checkConsistencyWithStrictUnits, so that callers can decide as to the status of unit issues. 

The PR is written so that the behavior with it is the same as before: 

* callers have to explicitly enable `StrictUnitsCheckON` with `setConsistencyChecks` for it to be run with `checkConsistency`
* the test shows the behavior of both modes

## Motivation and Context
fixes #385 

## Types of changes

- [x] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/SBMLDocument.cpp
+++ b/src/sbml/SBMLDocument.cpp
@@ -730,6 +730,12 @@ SBMLDocument::checkConsistency ()
 }
 
 
+unsigned int
+SBMLDocument::checkConsistencyWithStrictUnits ()
+{
+  return checkConsistencyWithStrictUnits(LIBSBML_OVERRIDE_ERROR);
+}
+
 /*
  * Performs a set of semantic consistency checks on the document.  Query
  * the results by calling getNumErrors() and getError().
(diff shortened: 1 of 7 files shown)


skeating (approved):

matthiaskoenig:
Thanks. This is great.

PR #412, opened 2024-11-21
Fix curve child listOfElements

luciansmith:
In libsbml, a 'listOfElements' from the spec is called a 'listOfCurveElements' because of unique naming constraints.  However, when reading in an SBML file, the list was requried to be called 'listOfCurveElements' instead of 'listOfElements'.  This meant that when libsbml wrote out a file with a 'listOfElements' child of a 'RenderCurve', it couldn't be read back in again.

Instead of changing the required name from 'listOfCurveElements' to 'listOfElements', this change allows both, for backwards compatibility.

I added a test file for validation, but it isn't super compact; it's just the file we had created when we found the bug in the first place.

### DIFF
--- a/src/sbml/packages/render/sbml/RenderCurve.cpp
+++ b/src/sbml/packages/render/sbml/RenderCurve.cpp
@@ -1142,7 +1142,7 @@ RenderCurve::createObject(XMLInputStream& stream)
 
   const std::string& name = stream.peek().getName();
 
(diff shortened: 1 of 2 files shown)


fbergmann (approved):

luciansmith:
Thanks for this and the other merges!  Happily this means that roadrunner is back to using the develop branch of libsbml instead of its own offshoot.

Issue #416, opened 2025-03-25
Enhancement:  add markdown conversion for setting/getting 'notes'


luciansmith:
In SBML, the 'notes' object has to contain well-formed HTML.  This is, of course, a pain.

What if we looped in a markdown converter so that the user could get/set notes with markdown-style text, instead?

I was considering doing this for Antimony directly, but then thought the community might also benefit from this addition.

Tagging @matthiaskoenig @fbergmann @skeating just for opinions; if we thought this was a good idea, I'd work on it myself.

fbergmann:
sounds like a good idea, as long as there are no dependencies that would make it hard that'd be fine with me. Googling for header only markdown library i found <https://github.com/progsource/maddy> that could work

luciansmith:
Oh, nice, I hadn't found that one.  I don't suppose you know of one that does the reverse, from HTML to markdown?  Those seem to be more rare.

matthiaskoenig:
Very good idea. 
I am basically doing the same in sbmlutils and applying a MD -> HTML conversion before writing the notes, but the native notes format is markdown for me.

luciansmith:
Good to know!  Do you have a C/C++ HTML -> MD converter that you like?

fbergmann:
HTML to MD will definitely be more tricky, as it will be lossy. if that i fine there seem to be MIT licensed code for that too: <https://github.com/kstenschke/html2md> or <https://github.com/tim-gromeyer/html2md/>

luciansmith:
Looks like the first stopped getting updates in 2020, but the second is not only still active, but it only consists of two files, so that seems perfect.  Thank you for finding them!

And yes, I'm fine with HTML to MD being lossy, but I absolutely need to be able to round-trip things for my own project, and I assume that others will need that as well.  I think as long as we mention this in the documentation for the function, we should be OK.

matthiaskoenig:
Some thoughts:

1. You could also do some hacks to avoid roundtripping. I.e. just write the escaped markdown string in
```
<pre><code>
escaped markdown here
</code></pre>
```
And then unescape it if you load it back. You would not have nice rendering of the comments but tools could easily handle that if wanted. 

2. You have to be aware that html can be used in markdown. I.e. it is valid to have html tags in markdown. So you have probably have to deal with this or just say: simple markdown without html allowed only. Especially for image resizing and more control over displays of images I use an `<image>` tag in my markdown.
So you will have HTML in the HTML which cannot be backconverted.

[...]

PR #421, opened 2025-04-04
Add html/markdown translators.

luciansmith:
Add the ability to set and get SBase notes and Constraint messages in the 'markdown' format, as translated md->html using maddy, and html->md using html2md.

I made no attempt to save the raw markdown for now; the functions simply translate to and from each other.  If the user wants to store the raw markdown as an annotation, they'll need to do that themselves (at least for now).

Function names are

setNotesFromMarkdown
setMessageFromMarkdown

getNotesMarkdown
getMessageMarkdown

There wasn't a 'setNotesString' function, hence the 'From'.  But the names are negotiable if someone thinks they should be changed.

From #416.

### DIFF
--- a/src/sbml/maddy/CMakeLists.txt
+++ b/src/sbml/maddy/CMakeLists.txt
@@ -1,0 +1,61 @@
+# This project is licensed under the MIT license. For more information see the
+# LICENSE file.
+
+cmake_minimum_required(VERSION 3.25)
+
+project(maddy)
+
+# ------------------------------------------------------------------------------
+
+set(CMAKE_BINARY_DIR ${CMAKE_CURRENT_SOURCE_DIR}/build)
+set(EXECUTABLE_OUTPUT_PATH ${CMAKE_BINARY_DIR})
+set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR})
+
+# ------------------------------------------------------------------------------
+
+option(MADDY_BUILD_WITH_TESTS "enable building tests - does not work with zip download" OFF)
+
+if(${MADDY_BUILD_WITH_TESTS})
+  enable_testing()
+endif()
+
+option(MADDY_CREATE_PACKAGE "create a package for a version release" OFF)
+
+# ------------------------------------------------------------------------------
+
+set(MADDY_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/include)
+
+# ------------------------------------------------------------------------------
+
+if(NOT DEFINED CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
+if(${CMAKE_CXX_STANDARD} LESS 14)
+  message(FATAL_ERROR "maddy requires >=C++14")
+endif()
+
+# ------------------------------------------------------------------------------
+
+add_library(maddy INTERFACE)
+target_include_directories(maddy INTERFACE
+  ${MADDY_INCLUDE_DIR}
+)
+
+# ------------------------------------------------------------------------------
+
+if(${MADDY_BUILD_WITH_TESTS})
+  add_subdirectory(tests)
+endif()
+
+# ------------------------------------------------------------------------------
+
+if(${MADDY_CREATE_PACKAGE})
+  set(MADDY_PACKAGE_FILES include/ CMakeLists.txt LICENSE)
+  add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/${PROJECT_NAME}-src.zip
+    COMMAND ${CMAKE_COMMAND} -E tar c ${CMAKE_BINARY_DIR}/${PROJECT_NAME}-src.zip --format=zip -- ${MADDY_PACKAGE_FILES}
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    DEPENDS ${MADDY_PACKAGE_FILES})
+  add_custom_target(${PROJECT_NAME}_package DEPENDS ${CMAKE_BINARY_DIR}/${PROJECT_NAME}-src.zip)
+endif()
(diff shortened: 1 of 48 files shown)


luciansmith:
OK, so maddy (at least) requires C++11, which we are not set up to handle.  Should we?  It's 15 years old; it seems safe...

luciansmith:
Update: it's just some of our github actions that don't use C++11.  In general, the other actions compile and run fine.  So if we can update the github actions, we'd be OK?

src/sbml/maddy/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+# This project is licensed under the MIT license. For more information see the
+# LICENSE file.
+
+cmake_minimum_required(VERSION 3.25)
```
fbergmann, line 4: since maddy requries cmake 3.25 it would be best that all the other requirements are also upgraded currently we still have a wider range that we allow compilation with cmake_minimum_required(VERSION 2.8.4...3.19). We might also have to upgrade the c++ standard to c++14 which is fine with me.

fbergmann (changes requested):
we need to upgrade the cmake build to require c++14 everywhere at least and a more recent cmake version.

fbergmann:
> Update: it's just some of our github actions that don't use C++11. In general, the other actions compile and run fine. So if we can update the github actions, we'd be OK?

yes we ought to update the actions and the cmake scripts to ensure that everyone is made aware.

fbergmann:
@luciansmith i actually would prefer if we had a general static util methods something like `util_html_to_markdown` / `util_markdown_to_html`. With c overloads. That would be useful from all other projects using libsbml. We'd then be just calling those methods everywhere. This would reduce the places where we need the new includes, and would have the concatenation also happening only once. Thanks!

src/sbml/maddy/CMakeLists.txt
luciansmith: I included the maddy CMakeLists file just because it was in the original, but we don't actually use it.  If this is confusing, I could just remove the file entirely?

luciansmith:
It looks like you can't actually set a minimum C++ version in CMake; you can only set a particular C++ version ('use C++ 11 always'), unless you want to check the flags manually ('make sure the user has '-std=c++11' or '-std=c++14' or ...'). See https://stackoverflow.com/questions/48148275/inducing-minimal-c-standard-version-in-cmake

So for now, I just changed the github action to use 11 instead of 98.  (We didn't need C++14 yet, just 11.)  Is that reasonable for now?

In other news, I added the functions you requested, but I wasn't sure about the names.  For now, I went with

util_html_to_markdown
util_markdown_to_html
util_html_to_markdown_c
util_markdown_to_html_c

The latter two feel clunky to me, but I couldn't think of anything better to name the C versions of those functions.  Happy to take other suggestions, if you have them.

luciansmith:
I was wrong!  You can indeed tell CMake to look for features you need with target_compile_features.  Added and tested.  It actually works even with the original github actions that try to use c++98; it adds the c++11 flag automatically.

src/sbml/maddy/CMakeLists.txt
fbergmann: ok, as long as we dont require it that is fine then. Might make sense to remove it, but in any case that does not matter then. Thanks!

fbergmann (approved):
@luciansmith thanks for this, this seems to work fine for me. I've added a bit to expose the util file to the bindings. With that i'm able to call it from python

luciansmith:
Whoops--exposing the util functions actually created a bunch of other errors as well in the rapid build; it's not just the ubuntu failed builds.  Re-opening this to see if it can be fixed.

luciansmith:
For the full build, on Windows:
2025-04-10T13:50:58.1775605Z csharp-files\libsbml.cs(531,97): error CS1503: Argument 1: cannot convert from 'System.Runtime.InteropServices.HandleRef' to 'string'

On Ubuntu:
2025-04-10T13:37:44.6406371Z /home/runner/work/libsbml/build/src/bindings/java/libsbml_wrap.cpp:3118:8: error: invalid array assignment
2025-04-10T13:37:44.6407247Z  3118 |   arg4 = *argp4;
2025-04-10T13:37:44.6407578Z       |   ~~~~~^~~~~~~~

and:

2025-04-10T13:37:44.1821060Z /home/runner/work/libsbml/build/src/bindings/csharp/libsbml_wrap.cpp:2094:8: error: invalid array assignment
2025-04-10T13:37:44.1821568Z  2094 |   arg4 = *argp4;
2025-04-10T13:37:44.1821768Z       |   ~~~~~^~~~~~~~

For the 'rapid' build, on windows:

csharp-files\libsbml.cs(531,97): error CS1503: Argument 1: cannot convert from 'System.Runtime.InteropServices.HandleRef' to 'string'

On ubuntu:
2025-04-10T13:35:02.5332898Z /home/runner/work/libsbml/build/src/bindings/r/libsbml_wrap.cpp: In function ‘SEXPREC* R_swig_c_locale_vsnprintf(SEXP, SEXP, SEXP, SEXP, SEXP)’:
2025-04-10T13:35:02.5334317Z /home/runner/work/libsbml/build/src/bindings/r/libsbml_wrap.cpp:5655:14: error: invalid array assignment
2025-04-10T13:35:02.5335180Z  5655 |         arg4 = *(reinterpret_cast< va_list * >(argp4));
2025-04-10T13:35:02.5335646Z       |         ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

On manylinux:
/__w/libsbml/build/src/bindings/java/libsbml_wrap.cpp: In function ‘jint Java_org_sbml_libsbml_libsbmlJNI_c_1locale_1vsnprintf(JNIEnv*, jclass, jstring, jlong, jstring, jlong)’:
/__w/libsbml/build/src/bindings/java/libsbml_wrap.cpp:3973:8: error: invalid array assignment
 3973 |   arg4 = *argp4;
      |   ~~~~~^~~~~~~~

luciansmith:
[I can't actually re-open this; addressed in other PRs instead.]

PR #426, opened 2025-04-10
Use latest C++ compilers.

luciansmith:
Change the 'rapid build' github action to use c++26 when possible.  This should allow us to catch any changes (dropped constructs, etc.) sooner rather than later.

Worked out of the box, so no problems as of now.

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -30,7 +30,7 @@ jobs:
         package_option: ["-DWITH_ALL_PACKAGES=ON"]
         language_bindings:
           [ "-DWITH_JAVA=True -DWITH_CSHARP=True"]
-        cpp_standard: [98, 20]
+        cpp_standard: [98, 26]
     runs-on: ${{ matrix.platform }}
 
     steps:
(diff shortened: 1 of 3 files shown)


luciansmith:
Unrelatedly, 'yum' stopped working on the quay build, because it's looking for the mirror URLs:

```
http://westus2.azure.repo.almalinux.org/almalinux/8.10/BaseOS/x86_64/os/Packages/c/crypto-policies-scripts-20230731-1.git3177e06.el8.noarch.rpm
```

but there's no (or no longer?) any 'c/' subdirectory; it's actually just available at:

```
http://westus2.azure.repo.almalinux.org/almalinux/8.10/BaseOS/x86_64/os/Packages/crypto-policies-scripts-20230731-1.git3177e06.el8.noarch.rpm
```

luciansmith:
Yum works again; builds now succeeding once more.

fbergmann:
that works for me ... though we should not have the  changes to libsbml.i / libsbml.h here. After that i think it is good to merge.

luciansmith:
Done!

fbergmann (approved):
awesome thank you so much

Issue #359, opened 2024-01-12
Update SBO


luciansmith:
At least for the online validator, I get this error:

Warning[ Line 2](https://synonym.caltech.edu/validator_servlet/section-print-result.jsp#line-2) Column 34: (SBML Validation Rule #10719) The value of the 'sboTerm' attribute on the <sbml> object is expected to be an SBO identifier (http://www.biomodels.net/SBO/) referring to a modeling framework (i.e., terms derived from SBO:0000004, "modelling framework"). Reference: L3V2 Section 5 SBO:0000675 does not derive from the modelling framework branch.
<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" sboTerm="SBO:0000675" level="3" version="2">

However, 675 does indeed inherit from 4, but it was added more recently.  We just need to pull in the latest SBO ontology.

fbergmann:
turn out newer sbo terms have also changed so that `99702-fail-01-01-10703.xml` no longer fails, looking for the right term to use instead

fbergmann:
As it turns out the new obo export no longer marks terms as obsolete: <https://github.com/EBI-BioModels/SBO/issues/9>

fbergmann:
this has now been merged with the manual changes. So currently running `update_sbo` would break it again.

PR #433, opened 2025-05-01
#429 : ensure that annotations / notes of kvps are written

### LINKED ISSUE #429: FBC3 KeyValuePair annotations + metaid
Hi,

I recently picked up the implementation of SBML L3 FBC v3 in cobrapy (see https://github.com/opencobra/cobrapy/pull/1440). All of the specification implemented so far was nicely available in libsbml, so thank you for that! The only issue I ran into is with KeyValuePair objects:
The issue is with `fbc-21501` and `fbc-21502` of the spec, which state that KeyValuePair can have sboTerm and metaid attributes, and an annotation subobject. This is indeed implemented in the libsbml code by having KeyValuePair inherit from SBase. However, the code to convert this into XML is not present. I think the issue lies within the functions at `FbcSBasePlugin::writeKeyValuePairsAnnotation` and `KeyValuePair::toXML()`. The latter indeed seems to lack any call to export SBase attributes etc., whilst the first lacks the ability to write the attributes metaid and sboTerm, as per `fbc-21509` (we're not using `fbc-21509` in cobrapy, so I do not really care about that, just noticed it).

The following code illustrates the issue:

```python
import libsbml
from logging import getLogger
logger = getLogger(__name__)

def check(value: int, message: str) -> bool:
    """Check the libsbml return value and prints message if something happened.

    If 'value' is None, prints an error message constructed using
      'message' and then exits with status code 1. If 'value' is an integer,
      it assumes it is a libSBML return status code. If the code value is
      LIBSBML_OPERATION_SUCCESS, returns without further action; if it is not,
      prints an error message constructed using 'message' along with text from
      libSBML explaining the meaning of the code, and exits with status code 1.
    """
    valid = True
    if value is None:
        logger.error(f"Error: LibSBML returned a null value trying to <{message}>.")
        valid = False
    elif isinstance(value, int):
        if value != libsbml.LIBSBML_OPERATION_SUCCESS:
            logger.error(f"Error encountered trying to '{message}'.")
            logger.error(
                f"LibSBML returned error code {str(value)}: "
                f"{libsbml.OperationReturnValue_toString(value).strip()}"
            )
            valid = False

    return valid


TEST_SBML = """<?xml version="1.0" encoding="UTF-8"?>
<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" xmlns:fbc="http://www.sbml.org/sbml/level3/version1/fbc/version3" level="3" version="2" fbc:required="false">
<model>
    <annotation>
        <listOfKeyValuePairs xmlns="http://sbml.org/fbc/keyvaluepair">
            <keyValuePair metaid="meta_kvp_1" id="kvp_1" key="cobra_test" value="true">
                <annotation>
                    <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:bqmodel="http://biomodels.net/model-qualifiers/" xmlns:bqbiol="http://biomodels.net/biology-qualifiers/">
                        <rdf:Description rdf:about="#meta_kvp_1">
                            <bqbiol:is>
                                <rdf:Bag>
                                    <rdf:li rdf:resource="http://identifiers.org/bigg.model/e_coli_core" />
                                </rdf:Bag>
                            </bqbiol:is>
                        </rdf:Description>
                    </rdf:RDF>
                </annotation>
            </keyValuePair>
        </listOfKeyValuePairs>
    </annotation>
</model>
</sbml>"""


def test_read():
    print("Reading nested metadata example.")
    doc: libsbml.SBMLDocument = libsbml.readSBMLFromString(TEST_SBML)

    model: "libsbml.Model" = doc.getModel()
    model_fbc: "libsbml.FbcModelPlugin" = model.getPlugin("fbc")

    for kvp in model_fbc.getListOfKeyValuePairs():
        cvterms = kvp.getCVTerms()
        for cv in cvterms:
            print(f"KeyValuePair->CVTerm->Resource[0]: {cv.getResourceURI(0)}")


def test_write():
    print("# Writing nested metadata example.")
    # create document with new fbc version 3 namespace
    sbmlns: libsbml.SBMLNamespaces = libsbml.SBMLNamespaces(3, 2)
    sbmlns.addPkgNamespace("fbc", 3)
    doc: libsbml.SBMLDocument = libsbml.SBMLDocument(sbmlns)
    doc_fbc: libsbml.FbcSBMLDocumentPlugin = doc.getPlugin("fbc")
    doc_fbc.setRequired(False)

    model: libsbml.Model = doc.createModel()
    model_fbc: libsbml.FbcModelPlugin = model.getPlugin("fbc")
    model.setMetaId("meta_model_e_coli_core")

    kvp: libsbml.KeyValuePair = model_fbc.createKeyValuePair()
    kvp.setKey("cobra_test")
    kvp.setValue("true")
    kvp.setId("kvp_1")
    check(kvp.setMetaId("meta_kvp_1"), "Setting metaid")

    cv: "libsbml.CVTerm" = libsbml.CVTerm()
    check(cv.setQualifierType(libsbml.BIOLOGICAL_QUALIFIER), "Set qualifier type")
    check(cv.setBiologicalQualifierType(libsbml.BQB_IS), "Set qualifier")
    check(
        cv.addResource("http://identifiers.org/bigg.model/e_coli_core"), "Add resource"
    )

    # check(model.addCVTerm(cv, newBag=True), "Adding CVTerm to model")
    check(kvp.addCVTerm(cv, newBag=True), "Adding cvterm")
    cvterms_check = kvp.getCVTerms()
    for cv_check in cvterms_check:
        print(f"KeyValuePair->CVTerm->Resource[0]: {cv_check.getResourceURI(0)}")

    sbml_str: str = libsbml.writeSBMLToString(doc)
    print("#####")
    print(sbml_str)
    print("#####")


if __name__ == "__main__":
    test_read()
    test_write()
```

The output of this is:
```
Reading nested metadata example.
KeyValuePair->CVTerm->Resource[0]: http://identifiers.org/bigg.model/e_coli_core
# Writing nested metadata example.
KeyValuePair->CVTerm->Resource[0]: http://identifiers.org/bigg.model/e_coli_core
#####
<sbml xmlns="http://www.sbml.org/sbml/level3/version2/core" xmlns:fbc="http://www.sbml.org/sbml/level3/version1/fbc/version3" level="3" version="2" fbc:required="false">
  <model metaid="meta_model_e_coli_core">
    <annotation>
      <listOfKeyValuePairs xmlns="http://sbml.org/fbc/keyvaluepair">
        <keyValuePair id="kvp_1" key="cobra_test" value="true"/>
      </listOfKeyValuePairs>
    </annotation>
  </model>
</sbml>

#####
```

This shows that setting annotations (cvterms) to a KeyValuePair works fine and reading it from an SBML file also works. It's just writing them to xml that is missing.
Hope you can help with this, let me know if you need more info.

fbergmann:
## Description
this pr syncs annotations before writing. That way potential cvterms are written out. I still would not recommend using kvps that way, but it is valid as far as SBML is concerned.

## Motivation and Context
fixes #429 

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


bgoli:
A general comment on the KVP annotations issue. Are they necessary, AFAIK KVP's don't inherit from SBase and I'm not at all sure if we want KVP's with annotations as it breaks the whole "easily to read" thing  ... or am I just not seeing this properly?

fbergmann:
They are not necessary as such. And indeed in the spec they dont inherit from SBase. In the libSBML implementation they did inherit from it though. And so the possibility in libSBML to set / get them was always there just not implemented. 

Now someone found a use for them in the issue raised. Like i mentioned in the issue there, i would rather see the CVTerm on the SBase element itself rather than on the KVP.

fbergmann:
@bgoli thinking about it further ... it is right that the KVP is not an sbase element as such, after all it lives in an annotation. That on the other hand means that anything goes, we dont enforce anything in annotations, as such there might as well be an annotation on the annotation. If anything we probably ought to add a best practice blurb into the spec, stating that it would be best to annotate the object as such, not one KVP.

bgoli:
Thanks that clarifies  things for me enough to move ahead.

bgoli (approved):
Thanks @fbergmann looks good to me.

fbergmann:
this now also fixes #359

PR #435, opened 2025-05-02
Update Maddy to not convert internal single underscores.

luciansmith:
Still has problems with internal double-underscores, but at least those are less common.

I've been going through all the 'notes' in Biomodels models, checking to see if they roundtrip through the new markdown parsers.  Many don't!  I've been submitting issue requests and fixes to maddy and to html2md.  This is a pretty large fix, as there are many URLs with internal underscores in them that maddy was converting to `<em>`.

However, there are other problems; if you have to release now, we do indeed do what we claim (use those two libraries to convert markdown); it's just that those libraries have issues.

### DIFF
--- a/src/sbml/maddy/emphasizedparser.h
+++ b/src/sbml/maddy/emphasizedparser.h
@@ -40,8 +40,10 @@ public:
    */
   void Parse(std::string& line) override
   {
+    // Modifed from previous version, with help from
+    // https://stackoverflow.com/questions/61346949/regex-for-markdown-emphasis
     static std::regex re(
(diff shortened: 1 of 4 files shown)


fbergmann:
@luciansmith a release right now would at least help with the issues of broken python3.13 bindings, FBC issues, and the other fixes over the past year or so. We can certainly have another release whenever the MD support has increased further. But i'm happy to pull the 5.20.5 draft release for now.

luciansmith:
If a release means that other things get fixed, that's fine!  If this fix could get squeezed in, that'd be great, but it's not crucial.  Since there will be other fixes in the pipeline, if this fix gets bundled with those, that's fine.

fbergmann (approved):
@luciansmith this looks good to me, shall i merge it in and release?

luciansmith:
Yeah, let's do it.  Thank you!

PR #437, opened 2025-05-07
Test whether big/little endian using newer constructs.

luciansmith:
For some reason, Windows WSL/Ubuntu didn't work with the older method (which I believe is now deprecated); use the newer 'CMAKE_C_BYTE_ORDER' instead.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -147,12 +147,9 @@ check_include_files (stdlib.h STDC_HEADERS)
 check_include_files (string.h STDC_HEADERS)
 
 set(WORDS_BIGENDIAN)
-if (UNIX)
-  include (TestBigEndian)
-  test_big_endian(WORDS_BIGENDIAN)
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -147,12 +147,9 @@ check_include_files (stdlib.h STDC_HEADERS)
 check_include_files (string.h STDC_HEADERS)
 
 set(WORDS_BIGENDIAN)
-if (UNIX)
-  include (TestBigEndian)
-  test_big_endian(WORDS_BIGENDIAN)
+if(CMAKE_C_BYTE_ORDER STREQUAL "BIG_ENDIAN")
```
fbergmann, line 150: @luciansmith CMAKE_C_BYTE_ORDER was added in cmake 3.20, it would be good if we could leave the other statment in there for previous versions of cmake

fbergmann (approved):

PR #464, opened 2025-12-28
Sort multiple compartments

skeating:
## Description

Ratel to reaction conversion will only fail with multiple compartments if the species referred to in a given rate rule come from different compartments
## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing


fbergmann (approved):
Thank you so much, I've tested with BioMD 299, and it worked as expected. 

While I think that `speciesFromMultipleCompartmentsInSameRateRule` looks way more complicated than it should (6 levels of indentation), it's fine. No need to change to a hash_map for simpler access.

PR #472, opened 2026-03-05
Try upgrading to swig 4.4.1

luciansmith:
Might help with https://github.com/sbmlteam/python-libsbml/issues/40

### DIFF
--- a/.github/workflows/brief.yml
+++ b/.github/workflows/brief.yml
@@ -71,9 +71,9 @@ jobs:
         if: matrix.platform == 'windows-latest'
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


fbergmann:
@luciansmith the python bindings are not created out of this repo (rather from python-libsbml), and I'm taking care of this. I agree that the build scripts in this repo should change at some point. But it should probably a bigger change than just the SWIG version 

Fun fact the SWIG version 3.0 that was being used here, did not have the issue reported

fbergmann:
@luciansmith the reason we left the test here at swig 3.0 is, that the C# and Java bindings would have to be modified to work with newer versions of SWIG. So this change would have to be part of this PR.

PR #474, opened 2026-03-09
Allow building zlib and libsbml at the same time.

luciansmith:
The check_library_exists function breaks my build when building both zlib and libsbml at once.  I get the error:

CMake Error at C:/Users/Lucian/Desktop/libroadrunner-deps-backup/build-debug/CMakeFiles/CMakeScratch/TryCompile-m25yy5/CMakeLists.txt:28 (target_link_libraries):
  Error evaluating generator expression:

    $<TARGET_LINKER_FILE:zlibstatic>

  No target "zlibstatic"

If I then hit compile again, it works!  But the first time, it fails, and it needs to work the first time for my github action.

This problem was obfuscated by it previously trying to assign the result to LIBZ_CMAKE_PATH, which already existed, so it... broke the check?  Maybe?  Anyway.

I don't think we need to check that the library exists anyway.  We don't do it for any of the XML libraries, and only do it for BZip.  Which would probably also break my build, but I'm not including bzip in my build at all.

### DIFF
--- a/CMakeModules/FindZLIB.cmake
+++ b/CMakeModules/FindZLIB.cmake
@@ -31,7 +31,7 @@ if (NOT (ZLIB_INCLUDE_DIR AND ZLIB_LIBRARY) OR NOT ZLIB_FOUND)
     find_path(ZLIB_INCLUDE_DIR zlib.h zlib/zlib.h
 	    PATHS $ENV{ZLIB_DIR}/include
 	          $ENV{ZLIB_DIR}
-              ${${_PROJECT_DEPENDENCY_DIR}}/include
+	          ${${_PROJECT_DEPENDENCY_DIR}}/include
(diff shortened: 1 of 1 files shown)


fbergmann (approved):
@luciansmith rather than leave the code there commented out i just disabled it, leaving people to enable it if they had a need. 

This should solve your original issue. 

Otherwise its fine to just remove the statements altogether, but then the line changing the path to the zlib library should also be changed. Oh and the error message was just wrong and did not actually print the name of the zlib library.

luciansmith:
Looks good to me!  Is there something more you wanted someone to do, cf "changing the path to the zlib library should also be changed"?  Or is this ready to merge?

fbergmann:
@luciansmith thats all included in my change so its ready to merge and i will

PR #475, opened 2026-03-10
Allow a model history parent to be set.

luciansmith:
Also, set the parent correctly when adding one to an existing element.

Found a bug in Antimony where I would copy a ModelHistory object from one document to another, and it came with its old parent, which would often be deleted.  However, there was no way for me to unset the parent of a ModelHistory, so this fixes that.

### DIFF
--- a/src/sbml/SBase.cpp
+++ b/src/sbml/SBase.cpp
@@ -2362,6 +2362,7 @@ SBase::setModelHistory(ModelHistory * history)
     {
       delete mHistory;
       mHistory = static_cast<ModelHistory*>(history->clone());
+      mHistory->setParentSBMLObject(this);
       mHistoryChanged = true;
       status = LIBSBML_OPERATION_SUCCESS;
     }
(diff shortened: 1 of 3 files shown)


fbergmann (changes requested):
@luciansmith great catch! I agree that the SBase change, definitely should go in. 

the new 'connectToParent' method, i'm not so sure about. We have the function to set the parent object. And ModelHistory is not an SBase derived class (or an SBML object), so I would find the method there confusing. Would you be fine with just the sbase change getting in?

luciansmith:
The other method is private, so I can't call it from Antimony.  Originally I just made it public, but looking at other classes, 'connectToParent' was the function we usually used.

I would be fine with either the current change, or the other parent-related functions being made public instead of protected!

fbergmann:
@luciansmith lets make the isSet/get/set/unset ParentObject methods public! That would be even better than the connectToParent

fbergmann (approved):
thanks a lot

PR #477, opened 2026-03-10
Remove warnings

luciansmith:
* sprintf->snprintf
* Addresses of actual objects are never NULL
* 'this' is never NULL
* Add default cases for ASTNode switch statements.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -48,6 +48,7 @@ foreach(example
     addModelHistory
     appendAnnotation
     callExternalValidator
+    checkNotesMarkdown
     convertSBML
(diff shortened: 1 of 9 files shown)


fbergmann (approved):
looks good to me though it seems odd, that we have different array lengths in our code base. 

and i'd much rather see the use of normal stl instead, like std::to_string rather than snprintf. 

and the group thing, since `isSetGroup` always return true, we really dont need to check for isSetGroup ever.

luciansmith:
Yeah, I left in the 'isSetGroup' function just in case we changed our implementation in the future.  Like, perhaps we'll want to check if anything is non-default in the 'group' object, and only write things out then?

PR #481, opened 2026-06-27
Fix #354: Remove empty RDF Description nodes after unsetModifiedDates

### LINKED ISSUE #354: Adding and removing modified times leaves detritus
Because of Reasons, I'm adding modified dates to a model, then replacing it with a new modified date.  So inbetween, I call 'unsetModifiedDates'.  This mostly works, but I end up with a bunch of empty RDF 'Description' objects:

```
        <rdf:Description rdf:about="#foo">
          <dcterms:modified rdf:parseType="Resource">
            <dcterms:W3CDTF>2019-07-29T10:53:09Z</dcterms:W3CDTF>
          </dcterms:modified>
        </rdf:Description>
        <rdf:Description rdf:about="#foo"/>
        <rdf:Description rdf:about="#foo"/>
        <rdf:Description rdf:about="#foo"/>
        <rdf:Description rdf:about="#foo"/>
        <rdf:Description rdf:about="#foo"/>
```

Can those be removed as well?  I couldn't figure out where they came from.

dyrpsf:
## Description
Updated the `SBase::reconstructRDFAnnotation()` cleanup logic in `src/sbml/SBase.cpp`. Added a routine at the end of the function to iterate backwards through the `RDF` node's children and delete any `<rdf:Description>` elements that have zero inner children. If the parent `<rdf:RDF>` block becomes completely empty as a result, it is also safely removed. 
I also added a unit test (`test_SBase_unsetModifiedDates_issue354`) in `src/sbml/test/TestSBase.cpp` to explicitly verify that no detritus is left behind when dates are removed.

## Motivation and Context
Previously, calling `unsetModifiedDates` would successfully remove the date nodes but leave behind empty `<rdf:Description rdf:about="#foo"/>` elements in the XML tree. Over multiple cycles, these empty tags would accumulate as detritus in the model's annotation. This change ensures the XML tree remains clean after node removal.
Fixes #354

## Types of changes
- [ ] Change in documentation

## Checklist:

## Testing
- [x] Testing is done automatically and codecov shows test coverage
- [ ] This cannot be tested automatically ```

### DIFF
--- a/src/sbml/SBase.cpp
+++ b/src/sbml/SBase.cpp
@@ -6470,6 +6470,32 @@ SBase::reconstructRDFAnnotation()
    }
 }
 
+if (mAnnotation != NULL)
+  {
+    int rdfIndex = mAnnotation->getIndex("RDF");
+    if (rdfIndex >= 0)
+    {
+      XMLNode& rdfNode = mAnnotation->getChild((unsigned int)rdfIndex);
+      
+      // Iterate backwards since we are removing items from the list
+      for (int i = (int)rdfNode.getNumChildren() - 1; i >= 0; i--)
+      {
+        XMLNode& child = rdfNode.getChild((unsigned int)i);
+        
+        // If the Description node has no inner children, delete it
+        if (child.getName() == "Description" && child.getNumChildren() == 0)
+        {
+          delete rdfNode.removeChild((unsigned int)i);
+        }
+      }
+      
+      // If the RDF node itself is now completely empty, delete it as well
+      if (rdfNode.getNumChildren() == 0)
+      {
+        delete mAnnotation->removeChild((unsigned int)rdfIndex);
+      }
+    }
+  }
 
   if (history != NULL) delete history;
   if (cvTerms != NULL) delete cvTerms;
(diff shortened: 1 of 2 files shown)


fbergmann (approved):
@dyrpsf thank you for this

dyrpsf:
Hi @fbergmann, thanks for the approval! 

I just pushed a quick update to fix the CI build failures. There was a syntax error in the unit test I added (mistakenly calling the date methods on `Model` instead of `ModelHistory`), which caused the test suite to fail compilation. 

The tests should compile cleanly and pass now. Let me know if anything else is needed before this can be merged!

dyrpsf:
Hi @fbergmann,

Apologies for the back-and-forth on the CI checks! I just pushed another small commit to fix the compilation error in the new test suite. 

I had mistakenly tried to call `unsetModifiedDates()` directly on the `ModelHistory` object in the test, which caused the build matrix to fail. I've corrected it to call the method properly on the `Model` object (via `SBase`), while ensuring the memory cleanups (`delete h;` and `delete d;`) remain intact.

The extensive CMake builds should compile cleanly and pass this time. Thanks again for your patience and the approval!

PR #482, opened 2026-07-11
Fix memory leak in GeneProductAssociation::getFbcAssociationFor.

luciansmith:
'addAssociation' cloned and added the object, but didn't delete it; now we call 'appendAndOwn' directly so a) we don't leak, and b) we're more efficient.

Still need to delete the object on failure.

### DIFF
--- a/src/sbml/packages/fbc/sbml/GeneProductAssociation.cpp
+++ b/src/sbml/packages/fbc/sbml/GeneProductAssociation.cpp
@@ -282,7 +282,8 @@ FbcAssociation* GeneProductAssociation::getFbcAssociationFor(const ASTNode* astn
         delete fbcand;
         return NULL;
       }
-      if (fbcand->addAssociation(child_fbca) != LIBSBML_OPERATION_SUCCESS) {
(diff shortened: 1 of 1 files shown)


fbergmann (approved):
while fbcand / fbcor ::addAssociation do not clone the association (they just call     `return mAssociations.append(fa);` after some checks), in case of failure the added association should still be deleted.

luciansmith:
'append' clones!  That's why it's different from 'appendAndOwn'.

```c++
  /**
   * Adds an item to the end of this ListOf's list of items.
   *
   * This method makes a clone of the @p item handed to it.  This means that
   * when the ListOf object is destroyed, the original items will not be
   * destroyed.  For a method with an alternative ownership behavior, see the
   * ListOf::appendAndOwn(@if java SBase@endif) method.
   *
   * @param item the item to be added to the list.
   *
   * @copydetails doc_returns_success_code
   * @li @sbmlconstant{LIBSBML_OPERATION_SUCCESS, OperationReturnValues_t}
   * @li @sbmlconstant{LIBSBML_INVALID_OBJECT, OperationReturnValues_t}
   *
   * @see appendAndOwn(SBase* disownedItem)
   * @see appendFrom(const ListOf* list)
   */
  int append (const SBase* item);
```

Thanks for the review!  Merging now.

PR #484, opened 2026-07-27
log species reference with hOSU and initialConcentration as implicit reference

### LINKED ISSUE #483: Validation rule 20906 doesn't catch hOSU species with initialConcentrations being used to set compartments
The attached model crashes roadrunner, because species S1 is set hasOnlySubstanceUnits=true, but also is set with an initialConcentration.  'S1' is then used to set the initial value for its compartment C.

In order to calculate S1's initial amount, it first needs to multiply by C, and in order to get C, it needs to get the initial amount of S1.

[substanceOnly_concentration_loop.xml](https://github.com/user-attachments/files/30143221/substanceOnly_concentration_loop.xml)

Because this is a circular definition, I think it should trigger validation rule 20906.

fbergmann:
## Description
added test for implicit cycle

## Motivation and Context

fixes #483

## Types of changes

- [ ] Change in documentation

## Checklist:

## Testing

### DIFF
--- a/src/sbml/validator/constraints/AssignmentCycles.cpp
+++ b/src/sbml/validator/constraints/AssignmentCycles.cpp
@@ -414,6 +414,11 @@ AssignmentCycles::checkForImplicitCompartmentReference(const Model& m)
         && s->getHasOnlySubstanceUnits() == false)
       {
         logImplicitReference(m, id1, s);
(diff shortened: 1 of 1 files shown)


luciansmith (approved):
Works for me.  Thank you!

PR #485, opened 2026-08-04
Fix ASTNode's copyNode for constants.

luciansmith:
This bug showed up when using the FunctionDefinitionConverter on SBML Test Suite models.  (Tests added.)


skeating (approved):
Works for me

fbergmann:
@luciansmith while i agree, that the name ought to be set. I just wonder, whether copyNode should not copy the 'name' from source. is it not the case, that an application could use other than those names for the constant types?

luciansmith:
Oh!  You remembered better than I; I had forgotten that the 'name' of a constant could also be set.  I'll make the change.

PR #486, opened 2026-08-21
Fix table round-tripping.

luciansmith:
The table from Biomodel 197 was being mangled both going and coming; this fixes both.  We should also push them upstream to the source libraries, too.

### DIFF
--- a/src/sbml/html2md/html2md.cpp
+++ b/src/sbml/html2md/html2md.cpp
@@ -152,6 +152,7 @@ Converter::Converter(const string *html, Options *options) : html_(*html) {
   tags_[kTagTableRow] = make_shared<Converter::TagTableRow>();
   tags_[kTagTableHeader] = make_shared<Converter::TagTableHeader>();
   tags_[kTagTableData] = make_shared<Converter::TagTableData>();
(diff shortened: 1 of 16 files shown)


fbergmann:
@luciansmith that looks great, if anything it is unfortunate, that the changes from:  <https://github.com/sbmlteam/libsbml/commit/92a3f5c16257818894a0791064854c64ade3b9d4> are undone with the commit. Those were needed last nov to compile.

luciansmith:
I can certainly add them back in, but it's weird that everything seems to be passing/compiling without the changes?  Did the failing compiler change, or something?

fbergmann:
> I can certainly add them back in, but it's weird that everything seems to be passing/compiling without the changes? Did the failing compiler change, or something?

it failed on another system, not the one we were testing under. Anyways it would be great to have it in, thanks

luciansmith:
Got it.  Do you remember which one?  I ask just because I'd like to include that information when I post a PR to the original library.

luciansmith:
Changes added!

fbergmann (approved):
@luciansmith thank you so much

PR #488, opened 2026-09-18
Update maddy to simplify regexes, so VS2026 won't complain.

luciansmith:
A round-tripped notes from Biomodels 12 was causing maddy to run into throws from the regexes when compiled with VS2026 (not 2022, nor from any other compiler).  The new regexes are simpler, and should work for more complicated strings, like the one found in biomodels 12.  Added a test with the Cursed String.


luciansmith:
Also needed a different fix in html2md: it was stripping the second character of UTF-8 non-breaking spaces.

fbergmann (approved):

